PR #149, opened 2015-01-02
148 non unicode strings

davidwtbuxton:
Think this is enough to force values to unicode when persisting. I ran into this bug when assigning the result of django.template.defaultfilters.slugify(..) to the field.

In theory the test should be OK. Not sure how one is supposed to run djangae's tests.


davidwtbuxton:
Dammit, wrong bug number in branch…

PR #159, opened 2015-01-12
Add unique constraint support for ListField and SetField

Kazade:
This adds support for using unique markers with ListField and SetField (e.g. a value can appear in the field for only one instance)


olasitarska:
Looks good! :sparkles:

PR #166, opened 2015-01-13
Allow gauth tests to run with custom user

adamalton:
See commit comments for detailed banter.


Kazade:
Nice one! Thanks :+1:

PR #175, opened 2015-01-14
Sandbox fixes env

chargrizzle:
Fixes the missing env variable that some appengine libraries (webapp) use to determine import paths


Kazade:
I thought you said that yaml wasn't on the path at this point? We can't rely on system wide yaml...

chargrizzle:
It is on the path after the path is set up above it above it before all the
imports

PR #179, opened 2015-01-15
Sandbox path sym

chargrizzle:
This makes sure requests that go through the WSGI handler have identical paths to when called through dev_appserver


Kazade:
I'm 99% sure this isn't right, and I'm fairly sure the current behaviour works, so I'm closing this. Let me know if you have any objections!

PR #185, opened 2015-01-15
import fix to make sure mapreduce is not a dep

chargrizzle:
mapreduce is an external library because the one bundled in the sdk is advised against, this just wraps a try around the import


olasitarska:
:sparkling_heart:

PR #189, opened 2015-01-16
Splitting special indexes that are too big for one field

olasitarska:
Uff


olasitarska:
Didn't notice that this breaks lots of tests...so I'm going to fix that and submit again ;)

PR #192, opened 2015-01-19
Sort the result programmatically when 'First ordering property must be the same as inequality filter property' exception occurs.

leekchan:
Hi.

<pre>
======================================================================
ERROR: test_object_creation (basic.tests.ModelTest)
----------------------------------------------------------------------
  File &quot;django_tests/basic/tests.py&quot;, line 459, in test_object_creation
    Article.objects.filter(id__lte=a4.id).delete()
  File &quot;djangae-testapp/lib/django/db/models/query.py&quot;, line 468, in delete
    collector.delete()
  File &quot;djangae-testapp/lib/django/db/models/deletion.py&quot;, line 260, in delete
    qs._raw_delete(using=self.using)
  File &quot;djangae-testapp/lib/django/db/models/query.py&quot;, line 479, in _raw_delete
    sql.DeleteQuery(self.model).delete_qs(self, using)
  File &quot;djangae-testapp/lib/django/db/models/sql/subqueries.py&quot;, line 85, in delete_qs
    self.get_compiler(using).execute_sql(None)
  File &quot;djangae-testapp/lib/django/db/models/sql/compiler.py&quot;, line 786, in execute_sql
    cursor.execute(sql, params)
  File &quot;djangae-testapp/lib/django/db/backends/util.py&quot;, line 53, in execute
    return self.cursor.execute(sql, params)
  File &quot;djangae-testapp/lib/djangae/db/backends/appengine/base.py&quot;, line 91, in execute
    self.rowcount = sql.execute()
  File &quot;djangae-testapp/lib/djangae/db/backends/appengine/commands.py&quot;, line 918, in execute
    self.select.execute()
  File &quot;djangae-testapp/lib/djangae/db/backends/appengine/commands.py&quot;, line 505, in execute
    self.gae_query = self._build_gae_query()
  File &quot;djangae-testapp/lib/djangae/db/backends/appengine/commands.py&quot;, line 671, in _build_gae_query
    query.Order(*ordering)
  File &quot;djangae-testapp/lib/google_appengine/google/appengine/api/datastore.py&quot;, line 1452, in Order
    (orderings[0][0], self.__inequality_prop))
BadArgumentError: First ordering property must be the same as inequality filter property, if specified for this query; received pub_date, expected __key__
</pre>

App Engine Datastore raises this exception when the first ordering property is not same as the inequality filter property. This limitation brokes the django test case and also causes failures when using those kinds of querysets. 

This patch makes those querysets work by sorting the result programmatically, but it may cause some cpu overheads. So I'm not sure this is the best solution. If you have a better idea, please comment. Then I'll change my codes.

Thanks.

### DIFF
--- a/djangae/db/backends/appengine/base.py
+++ b/djangae/db/backends/appengine/base.py
@@ -147,6 +147,14 @@ class Cursor(object):
             result.append(entity)
             i += 1
 
+        if hasattr(self.last_select_command, 'sort_ordering'):
+            for order, direction in reversed(self.last_select_command.sort_ordering):
+                if direction == 2:
+                    reverse = True
+                else:
+                    reverse = False
+                result.sort(key=lambda x: x[self.last_select_command.queried_fields.index(order)], reverse=reverse)
+            
         return result
 
     @property
--- a/djangae/db/backends/appengine/commands.py
+++ b/djangae/db/backends/appengine/commands.py
@@ -668,7 +668,11 @@ class SelectCommand(object):
                     query = datastore.MultiQuery(new_queries, ordering)
                 else:
                     query = queries[0]
-                    query.Order(*ordering)
+                    try:
+                        query.Order(*ordering)
+                    except Exception, exception:
+                        if str(exception).find('First ordering property must be the same as inequality filter property') >= 0:
+                            self.sort_ordering = ordering
         else:
             query.Order(*ordering)
 
--- a/djangae/db/backends/appengine/commands.py
+++ b/djangae/db/backends/appengine/commands.py
@@ -670,9 +670,11 @@ class SelectCommand(object):
                     query = queries[0]
                     try:
                         query.Order(*ordering)
-                    except Exception, exception:
-                        if str(exception).find('First ordering property must be the same as inequality filter property') >= 0:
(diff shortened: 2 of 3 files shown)


leekchan:
I fixed my previous commit not to swallow other exceptions.

Kazade:
So there are a couple of problems here. Firstly generally the approach doesn't work if the queryset is being limited (e.g. myqueryset[:10]) because you'll get 10 unsorted results from the datastore and then sort those, which won't give you the right result.

Also, regarding your implementation, checking the messaging of the exception is the wrong way to do this, we should instead determine before the query is built whether this is something we can do.

I'm not opposed to accepting a patch that sorts results when they are not limited and when this situation occurs (as sorting is a relatively fast operation compared to actually getting the data from the datastore) but we should log it somehow with some kind of configurable warning. A "slow query log" is something I've been meaning to setup anyway so perhaps this should wait until that is in place.

leekchan:
I appreciate your detailed feedback, which I agree in full. It is of my understanding that new approach should be implemented to facilitate the situation that I missed previously. 

As of now, I am closing this request for the time being while sourcing the better approach.

PR #215, opened 2015-01-23
Fixed error when indexed field value is None.

olasitarska:


Kazade:
This actually causes the world to explode, so I've reverted it!

olasitarska:
ugh ugh! I'll look on this later, sorry!

PR #216, opened 2015-01-26
Speed up 'get_field_from_column' by caching fields

leekchan:
This patch speed up 'get_field_from_column' by caching fields and executing loop only once for each model.

Thanks.


Kazade:
Perhaps it could just be as simple as just wrapping the function in djangae.utils.memoized?

leekchan:
I fixed it. Thanks.

olasitarska:
Thanks @leekchan! :zap:

PR #221, opened 2015-01-27
Uniquetool - A tool for adding in the uniqueness markers to a model that previously didn't have them

grzes:
A couple of django admin views for retrofitting the unique constraint markers when adding unique constraints to a model or migrating a model to Djangae.


Kazade:
One small nitpick, could you replace log_ids with log = RelatedSetField(...) ?

PR #251, opened 2015-02-23
Adding a retry function

grzes:
We had a retry until succesful, generalized it so we can make limited attempts to rerun a function.


adamalton:
Is doing `str(args)` and `str(kwargs)` not going to die when it hits unicode?

PR #270, opened 2015-03-03
Optionally get Google account credentials from ENV variable

igniteflow:
If $GOOGLE_AUTH is set (format: `email:password`) then use these to authenticate with the remote, otherwise prompt as normal

### DIFF
--- a/djangae/sandbox.py
+++ b/djangae/sandbox.py
@@ -120,11 +120,27 @@ def _local(devappserver2=None, configuration=None, options=None, wsgi_request_in
         os.environ = original_environ
 
 
+def get_auth_func():
+    logging.basicConfig(format='%(message)s')
+    logger = logging.getLogger(__name__)
+
+    GOOGLE_AUTH = os.environ.get('GOOGLE_AUTH')
+    if GOOGLE_AUTH:
+        try:
+            email, password = GOOGLE_AUTH.split(':')
+            logger.info(
+                'Using authentication credentials from environment variable GOOGLE_AUTH for account: {}'.format(email)
+            )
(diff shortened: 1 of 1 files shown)


adamalton:
I don't think we should be encouraging people to store their password in an environment variable, as that's just bad times.

I think there are 2 things we should do to improve the login experience for the remote API:

1. Make the username/password-based login store the login cookies (like djangoappengine used to).  These expire automatically.  See:
   * https://github.com/django-nonrel/djangoappengine/blob/master/djangoappengine/db/stubs.py#L20, and 
    * https://github.com/django-nonrel/djangoappengine/blob/master/djangoappengine/db/stubs.py#L115
2. Allow the `--oauth` flag to be passed into the `remote` command.  This can be done in a separate pull request though.

chargrizzle:
I looked into setting the --oauth2 flag for remote connections but unfortunatly it uses a different system from appcfg and does not support it :(

I believe the flag for the remote api login is --save-cookies? I think that's the best option

adamalton:
Ah, right `ConfigureRemoteApi` takes a `save_cookies` kwarg, so maybe we can pass that in rather than doing it via the `rpc_server_factory` thing which djangoappengine did.

I guess we could look for a `--save-cookies` flag, and only pass `save_cookies=True` if the flag is set.  But I never heard anyone complain about djangoappengine saving cookies by default, so maybe we shouldn't even both with the flag.

Kazade:
Closing this PR, we do need to fix up the remote sandbox. There has got to be a way to use oauth2 for this, might be worth looking through the appcfg code to figure out what they do.

Issue #277, opened 2015-03-12
djangae.contrib.gauth.models.Group causes errors if SQL is the default DB


adamalton:
In djangae.contrib.gauth.models we have a `Group` model which uses `ListField`.  If you import *anything* from djangae.contrib.gauth.models then the `Group` model gets registered by Django, and if SQL is your default DB then it doesn't know how to handle the `ListField` and so we get death.

Proposed solution:

* Take the models out of `djangae.contrib.gauth.models` and put them into 2 sub-apps:
    * `djangae.contrib.gauth.sql`
    * `djangae.contrib.gauth.datastore`
* Projects can then import the models which they want without importing the others as a side effect.
* For backwards compatibility, in `djangae.contrib.gauth.models` we should import the models from both the `sql` and the `datastore` sub apps, so that projects which are already importing from there don't break.
* We also need to make sure that `app_label` / `db_table` is correctly set on the models to ensure we don't break existing projects.

adamalton:
Ah, crap, those commits should have tagged #278.

PR #278, opened 2015-03-13
277 group causes error if sql is the default

sleepyjames:
This addresses #277 by splitting djangae.contrib.gauth into 2 separate applications. One for applications using the Datastore and one for applications primarily using CloudSQL (or other relational databases). I've prefixes 'gauth_' to the suggested applications names to avoid collisions (their quite generic names). I've also left references to the existing models for backwards compatibility.

I think there's some more work to be done here (refactoring the test for example) but this covers my use-case atm. Thoughts appreciated !


Kazade:
@adamalton could you review this please?

adamalton:
After my tweaks (which I accidentally tagged against #277), this now looks good.  My only lingering question is whether we should remove the `gauth_` prefix from the 2 sub-apps, so that we have `djangae.contrib.gauth.datastore` and `djangae.contrib.gauth.sql`.  Having the extra `gauth_` prefix niggles at me.

Having the prefix makes the app names less likely to clash with other apps, but given that the models have got `app_label = "djangae"` on them, I don't think it matters.

Thoughts @sleepyjames ?

Kazade:
I definitely think we should drop the gauth_ prefix FWIW :)

Issue #279, opened 2015-03-15
forms.fields.ListWidget doesn't work with set


pawciobiel:
forms.fields.ListWidget doesn't seam to work with sets.

adamalton:
Closed by #280

PR #280, opened 2015-03-15
fix forms.fields.ListWidget and ListFormField to work with model SetFiel...

pawciobiel:
...d

fix for https://github.com/potatolondon/djangae/issues/279


Kazade:
LGTM! Thanks!

PR #290, opened 2015-03-22
Adds travis icon indicating if the build is passing

pablorecio:


stucox:
shields.io can give you an SVG equivalent badge which looks nicer on high res screens and is accessible...

PR #298, opened 2015-03-25
Adding a test to verify cache is flushed only after the transaction comm...

grzes:
...its.

A parallel datastore op (like a deferred) is simulated with a non_transactional inline call.


Kazade:
Superseded by #300

PR #305, opened 2015-04-02
Make sure we set the default even on an update (Django doesn't do this)

Kazade:
Django doesn't call get_default() on an update, this is problematic on a schemaless database because we have no UPDATE query that could pre-populate it for us, so we always call get_default if the field is not-nullable, has a value of None, and has a default


Kazade:
@adamalton could you review this, this is going to get very annoying very quickly :)

PR #316, opened 2015-04-24
If a related set field is assigned a list of IDs, deal with it correctly...

Kazade:
... (allows RelatedSetField to be loaded in fixtures)

I'm not 100% convinced this is right, but it seems necessary :/


Kazade:
OK, after a bit of offline discussion, we've realized a better thing to do is to override the Python and JSON deserializers in settings and make them deal with RelatedSetField

PR #329, opened 2015-05-12
Add custom json serializer to handle sets

nealedj:
The dumpdata command was failing on RelatedSet fields with reverse relations because sets aren't serializable by default. Serializing RelatedSets directly will work because of the `value_to_string` method. However, this isn't called on the reverse end of the relationship. This adds a custom JSON serializer to simply handle all sets.

The loaddata command was failing for two reasons:

- The `handle_field` method on the serializer changes serialized lists to a string. This causes issues with ListField fields (such as user_permissions in contrib.gauth). `handle_field` is overridden so that ListFields get their values serialized as lists which can be set back during deserialization
- Deserializing was also failing for related sets when the reverse relation was set due to the setter expecting to always receive a list of models.


Kazade:
Sorry @nealedj , this is already handled. In your json you should just use:

thing: "[1, 2, 3]"

Note the quotes ^^. This is because Django uses to_python and value_to_string on custom fields for serialization and deserialization. A custom serializer isn't necessary and hopefully in Django 1.8(? maybe 1.9..) this situation will hopefully improve. But for now this is the right approach.

Sorry!

PR #330, opened 2015-05-13
Patch serialization to backport a serialization bugfix

nealedj:
Patch the `PythonSerializer.handle_fk_field` method to backport a fix added in Django 1.8. This was making `dumpdata` fail when it tried to serialize a `RelatedSetField` containing FKs.

Fix `IterableField` deserialization by adding string handing to the `to_python` method.

Fix case where deserialization to `RelatedSetField` of a long (999L) number would not parse.


nealedj:
Utterly mucked up this PR.

Opened a new one - https://github.com/potatolondon/djangae/pull/331

Issue #326, opened 2015-05-11
Add documentation for Read the Docs


Kazade:
In Markdown.. see here: https://read-the-docs.readthedocs.org/en/latest/getting_started.html#in-markdown

We should port the majority of the README to a docs folder, and then we can organize it into something sensible!

PR #344, opened 2015-05-19
Added real documentation OMG! 🔥

olasitarska:
Sooo, I really didn't want to write my talk tonight so to procrastinate a little bit I decided to make this. This PR should resolve issue #326 and basically all of our problems (obvs)

To test locally:

1. `pip install mkdocs`
2. `cd djangae`
3. `mkdocs serve`
4. Go to http://127.0.0.1:8000/

Known issues & questions:

1. I think mkdocs has a slightly different ReadTheDocs theme, hopefully this will update when we push it to RTD
2. I'm not sure if "Unique constraint tool" should be its own chapter, or maybe go into the "Unique Constraint Checking" chapter
3. I'm not sure if "Unique Constraint Checking" should be its own chapter, or maybe go into the "Database Backend" chapter
4. Does the order of chapters even make sense?

Yay! :tada:


Kazade:
You legend!!!

olasitarska:
@Kazade can you create a Potato account on Read the Docs in a free 5 minutes? https://read-the-docs.readthedocs.org/en/latest/getting_started.html#import-your-docs

PR #346, opened 2015-05-21
Adding an inline post to blobstore replacing the files api calls.

grzes:
A hacky replacement for the files api.


Kazade:
Looks good, could you throw a permission denied error if the following variable != '1':

os.environ['USER_IS_ADMIN']

That way if someone forgets to set it to login:admin we don't open a massive security hole :)

PR #360, opened 2015-06-05
RelatedList

chargrizzle:
This PR adds the ordered RelatedList to djangae, it also refactors RelatedSet into RelatedIter which makes everything a bit nicer.

RelatedList is an iterable field which guarantees order, and allows duplicate PKs.


Kazade:
Amazing work @chargrizzle !

PR #375, opened 2015-06-19
Iterable and related serialization

nealedj:
Basically getting `value_to_string` to do all of the serializing work in any `IterableField` and `RelatedIteratorField`. This prevents unicode strings from being serialized with a leading 'u' and longs from being serialized with a trailing 'L'.

Removed the JSON serializer patch as this only worked for Django <=1.7 on RelatedSetFields since they passed the set() value straight to the serializer. Django 1.8 runs the value through `value_to_string`. Have patched the Django 1.8 fix for earlier versions.


olasitarska:
@nealedj LukeB thinks this could have been fixed in master already.. Could you confirm?

nealedj:
Yes this looks like it's okay in master now.

PR #376, opened 2015-06-22
Handles TransformationError exception when getting blob's URL

pablorecio:
When uploading PDFs for instance, the Storage backend uses PIL's `get_serving_url` which raises a `TransformationError` exception as the file can't be transformed to image. This makes it a bit more fool-proof.


olasitarska:
Yay thanks! :penguin:

PR #380, opened 2015-06-24
Updated the MkDocs config from the deprecated format

d0ugal:


d0ugal:
This will fix the first build warning which can be seen here: https://readthedocs.org/builds/djangae/2991806/

(I think the second might be a MkDocs bug I recently fixed)

olasitarska:
Thanks @d0ugal! :sparkles:

PR #381, opened 2015-06-24
`django_ordering_comparison` can’t handle datetimes with null values

stucox:
Test to reproduce attached.

Will collaborate with @asendecka to add a fix to this PR.


olasitarska:
:banana: thanks!

PR #388, opened 2015-07-01
runserver supports the same commands than dev_appserver.py

pablorecio:
So, the idea here is to "cheat" Django's command's attributes adding the options from the sandbox as new elements in `option_list`.

I'm doing it in the `__new__` method so `option_list` can still be extended if someone wanted to extend the runserver command and add more options to it.

I wanted to add something to the docs, but I'm not sure in what section it would fit?

### DIFF
--- a/djangae/management/commands/runserver.py
+++ b/djangae/management/commands/runserver.py
@@ -23,24 +23,36 @@ class Command(BaseRunserverCommand):
     """
     # We use this list to prevent user using certain dev_appserver options that
     # might collide with some Django settings.
-    BLACKLISTED_DEV_APPSERVER_OPTIONS = [
-        'port',
-        'host',
-        'threadsafe_override',
-        'php_executable_path',
-        'php_remote_debugging',
-        'python_startup_script',
-        'python_startup_args',
-        'mysql_host',
-        'mysql_port',
-        'mysql_user',
-        'mysql_password',
-        'mysql_socket',
-        'smtp_host',
-        'smtp_user',
-        'smtp_password',
-        'smtp_allow_tls',
-        'automatic_restart',
+    WHITELISTED_DEV_APPSERVER_OPTIONS = [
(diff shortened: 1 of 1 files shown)


Kazade:
So, I've been meaning to look at this for some time, because although we want to support passing down *some* of the dev_appserver options, we don't want to do all of them as they conflict with Django options to runserver, off the top of my head:

 - --nothreading - This Django flag should be what the user calls, but our runserver command should pass the appropriate flags to dev_appserver to get single threaded behaviour
 - SQL settings - In Django, SQL settings are handled by settings.py (and --settings) we don't want to allow these to be passed down
 - Datastore/Blobstore etc. paths, these are handled internally and we probably don't want to expose them
 - Admin/API ports, we override these, we don't want users specifying them through runserver
 - SMTP settings, again, dealt with by Django settings
 - Automatic restart, is superseded by Django's --noreload

I think basically we need a whitelist of dev_appserver settings to expose in runserver which don't conflict with stuff we are doing in the sandbox, or with existing Django runserver flags. We can't just expose them all otherwise we're going to cause a *load* of weird bugs.

pablorecio:
I agree with some of those, but I think it'd be better to have a blacklist as I think most of the things make sense to have:

* Why don't we want to allow users to change the datastore / blobstore / search indexes / logs paths? I can't see a problem on that
* Admin / API ports - It's ok if we keep having the ones we have by default, but I think we should allow users to change those too.

I know that most of the things won't be used a lot, but I can't see any danger on users having access to those.

Kazade:
OK I guess, we'll have to check that doesn't break anything, but yeah I'm fine with that :)

pablorecio:
I'll amend the PR, and will check it more thoroughly. Cheers!

chargrizzle:
I think we should swap the blacklist for a whitelist, it's more explicit that way

PR #390, opened 2015-07-07
Docs tweaks

adamalton:


adamalton:
I rebased to fix my spelling of "pronunciation" in the commit message.  Otherwise the commit message would be nearly as ironic as the mistake it was fixing.

olasitarska:
:joy: I laughed at that comment.

PR #392, opened 2015-07-10
Docs on UniquenessMixin

ana-balica:
I am not sure what is the best place to mention about `UniquenessMixin`. The info about `unique` appears in `fields.md`, but the `uniquetool.md` seems to be focused on the uniqueness. If we provide one single example (without emphasizing the mixin) from the `fields.md`, it might be overlooked.

Sorry for the American English.

### DIFF
--- a/docs/uniquetool.md
+++ b/docs/uniquetool.md
@@ -41,3 +41,21 @@ support.
 ### Clean
 
 TODO: Docs needed.
+
+
+## UniquenessMixin
+
+If you want to mark a `ListField` or a `SetField` as unique or use `unique_together`
+Meta option with any of these fields, it is also **necessary** to use `UniquenessMixin`.:
+
+```
+from django.db import models
+from djangae.fields import SetField, ListField
+from djangae.db.constraints import UniquenessMixin
+
+
+class Princess(UniquenessMixin, models.Model):
+    name = models.CharField(max_length=255)
+    potential_princes = SetField(models.CharField(max_length=255), unique=True)
+    ancestors = ListField(models.CharField(max_length=255), unique=True)
+```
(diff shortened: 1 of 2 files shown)


adamalton:
From reading the code it looks like the `UniquenessMixin` is only necessary if you're putting `unique=True` on a `ListField` or `SetField`, right?

Kazade:
@adamalton yes that's right

ana-balica:
Updated the docs :cactus: 
Not sure if `UniquenessMixin` is required when using `unique_together` with any of these fields.

docs/uniquetool.md
```diff
@@ -30,3 +30,20 @@ problem and has to be resolved manually (change one of your instance's value so
...
+```
+from django.db import models
+from djangae import fields
+from djangae.db.constraints import UniquenessMixin
+
+
+class Princess(UniquenessMixin, models.Model):
+    name = models.CharField(max_length=255)
+    potential_princes = models.SetField(models.CharField(max_length=255), unique=True)
+    ancestors = models.ListView(models.CharField(max_length=255), unique=True)
```
olasitarska: should be `ListField`, instead of `ListView`.

also, we should use `fields.SetField` and `fields.ListField`, instead of `models`. Or even better, `from djangae.fields import SetField, ListField` and use them directly.

olasitarska:
It looks good now! I'm pretty sure mixin is also required when using unique_together on ListField or SetField.

There is one more thing to fix here, I commented above. Could you do it? :) :palm_tree:

docs/uniquetool.md
ana-balica: This is what happens when I don't test my code snippets. Thanks for noticing ;)

ana-balica:
Updated and rebased against latest master to spare you from conflict headaches.

olasitarska:
Thanks @ana-balica! :sparkles: :sparkling_heart:

PR #396, opened 2015-07-12
Allow usage of pk in djangae Paginator

### LINKED ISSUE #302: Allow usage of `pk` into djangae.contrib.pagination.Paginator
At the moment, it doesn't allow you to do something like:

    @paginated_model([('pk', )])
    class MyModel(models.Model):
        # things

As it doesn't detect `pk` as a valid field:

    PaginationOrderingRequired: No pagination ordering specified for ['pk']. Field required: pagination_pk

It does work with `id` though

jgasteiz:
Fixes #302


adamalton:
Looks good to me!

PR #397, opened 2015-07-13
fix typo

qiaoshun8888:
djan-gee -> djan-gae


adamalton:
No, that's a deliberate pronunciation guide!

Although it has been improved to make it less ambiguous in 31e952ab9978229c0e9683fa5ea248e8586c2c70

PR #402, opened 2015-07-22
Making populate() on ShardedCounterField idempotent/more transactional

### LINKED ISSUE #401: djangae.fields.counting.RelatedShardManager.populate should check existing number of shards
The `populate` method reloads the object inside a transaction in order to atomically update the `set` of shard PKs, but it checks the number of new shards which are needed **outside** of this transaction, which is silly.

adamalton:
This will close #401


adamalton:
This test error appears to be because the DB isn't being flushed between tests.

olasitarska:
Merged! :palm_tree:

PR #408, opened 2015-07-31
Adding a fix to @allow_modules so that it doesn't cause weirdness with os.environ

### LINKED ISSUE #405: sandbox.allow_modules causes death
TL;DR: `djangae.sandbox.allow_modules` screws with `os.environ`, so does the App Engine SDK, together these 2 monkey patches silently conflict with each other and cause things (usually transactions) to blow up.

For the full fairy tale, read on...

In `djangae.sanbox.allow_modules` we copy `os.environ`, mess with some stuff, and then do `os.environ = environ` (putting back our copy of the original).

Meanwhile in a land far far away, in the App Engine SDK, in `google/appengine/runtime/request_environment.py` there is a function called `PatchOsEnviron` which does this:

```
os_module.environ = RequestLocalEnviron(current_request)
```

i.e. it monkey patches `os.environ` so that it is separate for each request, even in a multi-threaded environment.

Then one day, having lived peacefully in separate lands, these 2 piece of code clashed in a violent battle.  Deep in the caves of `google.appengine.api.datastore` an innocent function called `__InitConnection` did this:

```
_ENV_KEY = '__DATASTORE_CONNECTION_INITIALIZED__'

def __InitConnection():
  """Internal method to make sure the connection state has been initialized."""
  if os.getenv(_ENV_KEY) and hasattr(_thread_local, 'connection_stack'):
    return
  _thread_local.connection_stack = [datastore_rpc.Connection(adapter=_adapter)]

  os.environ[_ENV_KEY] = '1'
```

this `__InitConnection` function gets used by `_SetConnection` which gets used by `RunInTransactionOptions` which gets used by `RunInTransaction`.  Which means that if you use `RunInTransaction` and then you use `djangae.sandbox.allow_modules` and then you use `RunInTransaction` again, you will get this:

```
File "google_appengine/google/appengine/api/datastore.py", line 420, in _PopConnection
    assert len(_thread_local.connection_stack) >= 2
AssertionError
```

adamalton:
Closes #405.

### DIFF
--- a/djangae/sandbox.py
+++ b/djangae/sandbox.py
@@ -384,7 +384,7 @@ def allow_modules(func, *args):
             mod.__dict__.update(_system.__dict__)
 
         # We have to maintain the environment, or bad things happen
-        os.environ = environ
+        os.environ = environ # This gets monkey patched by GAE
 
         try:
             return func(*args, **kwargs)
@@ -395,5 +395,7 @@ def allow_modules(func, *args):
             for mod in patch_modules:
                 _system = reload(mod)
                 mod.__dict__.update(_system.__dict__)
+            # Put the original os back, again
+            os.environ = environ
 
     return _wrapped


olasitarska:
Impressive investigation work @adamalton! :sparkles:

PR #413, opened 2015-08-03
Added RegexIndexer and IRegexIndexer

olasitarska:
:tada:


Kazade:
Amazing! Nice work!

adamalton:
```
class RegexIndexer(Indexer):

    def validate_can_be_indexed(self, value, negated):
        if negated:
            return False

        return isinstance(value, bool)
```

shouldn't that be `if negated and isinstance(value, (list, set))`?  Or am I misunderstanding it?  I thought that we could do negated regexes, just not for list/set fields?

PR #416, opened 2015-08-12
fix for `!=` queries: groupby requires a sorted list

martinogden:
This fixes the bug where `Model.objects.filter(**kwargs)` and `Model.objects.exclude(**kwargs)` are not always disjoint sets.


adamalton:
Those lines should have comments!

Seems to cause a couple of failures on Django 1.8.

Kazade:
The failing tests are actually caused by a mismatch between the Django tests we run (1.8.4) and the Django version on PyPi (1.8.3)

PR #422, opened 2015-08-19
Test fixes

martinogden:
* Workaround to get test models to work nicely with nose
 * Only run tests if functionality enabled in containing project


Kazade:
I'm pretty sure the skipIfCustomUser will always skip the test even in Djangae because we use a custom user model for gauth...

PR #430, opened 2015-08-26
Allow JSONField to store an OrderedDict

martinogden:


adamalton:
This feels like it's taking it into the realm of a PickleField rather than a JSONField, because an OrderedDict is not something which can be represented directly in JSON.

If we put this in then where do we stop?  Do we allow the JSONField to store `datetime` objects?  Do we allow it to store `set` objects, or `timedelta` objects, or a `namedtuple` or `Counter` or `<other-thing-from-collections-module>`?

I think that the OrderedDict stuff should probably be done in your application code rather than in the JSONField.  @Kazade agree?

Kazade:
Err.. All this does is allow retaining the order of dicts in the json when loaded/saved to the datastore. I need to do a proper review but I agree with the idea.

Json dicts do have an implicit order in the way they are defined, this just prevents losing that.

Kazade:
Actually, @martinogden can we just implement this as an "use_ordered_dict" argument to the field? The hook isn't going to work properly if it's anything other than OrderedDict so there's no point exposing it like that.

Add a use_ordered_dict kwarg, default it to False, make it all work with that.

PR #433, opened 2015-08-27
Gauth uses new app registry system

pablorecio:
This way, we connect and disconnect the permission migrations when the app is loaded, not during importation time, which is the Django-y way now.


pablorecio:
Wait, do not merge this yet. I think I've got to change something else...

pablorecio:
That looks more like it :smile:

PR #444, opened 2015-09-14
Make sure we store things in the context cache if we read them from m…

### LINKED ISSUE #443: We're not updating the context cache when we read from memcache
If we do a get by key (for example) and the element isn't in the context cache, we fallback to memcache if we aren't in a transaction. However, we never update the context cache so we continually hit memcache which is much slower.

This is a no-brainer massive speedup!

Kazade:
…emcache (which we only do if we're certain it's consistent) Fixes #443


olasitarska:
Looks good!

Issue #447, opened 2015-09-21
Resetting a counter breaks for negative counts.


armirusco:
Negative counts get handled in reset but then break when trying to do the actual increment. 

Fix on its way.

adamalton:
Closed via #448.

PR #448, opened 2015-09-21
Making sharded counters' `reset` work for negative counts. #447

armirusco:


adamalton:
Nice, thanks @armirusco !

PR #451, opened 2015-09-22
Adding default views to fire custom signals on an appengine module st…

### LINKED ISSUE #439: Handle _ah/start and _ah/stop
When using modules, App Engine will post to _ah/start and _ah/stop when starting and stopping instances. It would be good if we had default views for this which triggered custom django signals (e.g. instance_started and instance_stopped).

grzes:
…art/stop. Fixes #439.

### DIFF
--- a/djangae/core/signals.py
+++ b/djangae/core/signals.py
@@ -1,0 +1,6 @@
+from django.dispatch import Signal
+
+# Signals will only be fired when using GAE modules, not when autoscaling
+module_started = Signal(providing_args=['request'])
+module_stopped = Signal(providing_args=['request'])
+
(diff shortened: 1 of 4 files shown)


adamalton:
Thoughts:

* Should the signals be defined in `djangae.core.signals` (similar to `django.core.signals.request_started`)?
* Should they be called `module_started` to avoid confusion with "normal" instances?
* Do the `/_ah/start` and `/_ah/stop` paths need to be added to the handler [regexes in app.yaml](https://github.com/potatolondon/djangae/blob/master/testapp/app.yaml#L28)?
* Ditto for [app.yaml in djangae-scaffold](https://github.com/potatolondon/djangae-scaffold/blob/master/app.yaml#L9).

PR #456, opened 2015-09-26
fix django settings propagation which is passed as a command line argument

jbking:
This PR fixes switching django's settings module doesn't work now.

On local environment, djangae's runserver has capability of switching django's settings module by --settings. But its not passed into appropriate module's environment variable, even default django wsgi instance.

You can switch settings as it is. Addition, when you pass --${mod}-settings to the runserver command, its passed into the module's DJANGO_SETTINGS_MODULE environment variable.

### DIFF
--- a/djangae/core/management/__init__.py
+++ b/djangae/core/management/__init__.py
@@ -1,4 +1,5 @@
 import os
+import re
 import sys
 import argparse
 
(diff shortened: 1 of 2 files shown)


adamalton:
Thanks!

PR #459, opened 2015-09-30
Fix our handler to support files uploaded to GCS via blobstore.

grzes:
When using create_upload_url with a bucket name argument the files are uploaded to GCS, and an encoded blob key is passed to the app. Instead of a hash only the key has a form of: `encoded_gs_key:XXXXX` (where XXXX is the b64 encoded gcs filename). 

Pulling this simple change to our regex we can make our FileUploadHandler support gcs uploads through blobstore.


chargrizzle:
nice catch!

PR #460, opened 2015-10-06
Use oauth2 for the remote sandbox by default. We must store tokens in…

Kazade:
… a separate file as appcfg because we need different scopes


chargrizzle:
Fancy!

PR #461, opened 2015-10-07
Fixing edgecase where we would store a bad user object instead of swi…

grzes:
…tching id's

Users with uppercase emails were saved in 2 steps (first a lowercase version in the authentication code, and then a fixed version in our middleware). This caused migrating user id's break on users with uppercase letters.


Kazade:
LGTM!

PR #477, opened 2015-10-30
Only call unique_identifiers_from_entity if it was actually in the cache

### LINKED ISSUE #474: More efficient entity caching
Currently often entities are cached one by one, even when whole set is available.
That leads to situation when mc calls are actually slowing down `get` queries.
Most noticeable in `QueryByKeys`

grzes:
Fixes #474


grzes:
cache.get_many doesn't return None's for missing values! nevermind.

PR #478, opened 2015-11-03
Remove stray shims for Django 1.6 support.

timgraham:


testapp/install_deps.py
```diff
@@ -24,8 +24,6 @@
 
 DJANGO_VERSION = os.environ.get("DJANGO_VERSION", "1.7")
 NEXT_DJANGO_VERSION = {
```
timgraham: As far as I can tell, this variable isn't used in this project. Can it be removed? If not, it should be amended because 1.10 is the version that will follow 1.9.

Kazade:
Awesome! Thanks!

PR #480, opened 2015-11-06
models patch for GCS and custom app domain

trik:
models patch for GCS: prevent creation of foreign key constraints. now you can use syncdb and migrate on remote sandbox

custom app domain: added parameters to specify a custom gc domain (http or https)


olasitarska:
Sorry, this got out of date and needs to be rebased. :( Please re-open if you continue to work on it.

PR #484, opened 2015-11-17
Make the gauth middleware play nice with other authentication backend…

### LINKED ISSUE #483: gauth: gauth doesn't play nice with other backends
Imagine the situation where you want to support logging in with the Google users API, or a traditional username and password combination. In that case you'll enable both the gauth backend, and the standard Django ModelBackend. You'll also enable the gauth authentication middleware...

The problem is, the gauth middleware will log you out if you don't have a logged in Google user, so logging in with a username and password will immediately boot you back out again!

This just needs a bit of rejigging so that the only action that doesn't happen behind the existing backend check is to set request.user

Kazade:
…s. Fixes #483

### DIFF
--- a/djangae/contrib/gauth/middleware.py
+++ b/djangae/contrib/gauth/middleware.py
@@ -17,19 +17,12 @@ class AuthenticationMiddleware(DjangoMiddleware):
             if django_user:
                 login(request, django_user)
         else:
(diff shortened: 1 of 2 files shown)


adamalton:
Happy for this to be merged, but will let you review my changes first.

PR #486, opened 2015-11-18
Namespace support

Kazade:
This PR adds multiple datastore connections using namespaces. Note that like Django databases, cross-namespace foreign keys are not support.

This introduces a slight backwards incompatibility in the cache key naming (all cache keys now start with a namespace and a ":", if there is no namespace the cache key starts with a ":") so if you deploy to an existing version you will need to flush the cache.

### DIFF
--- a/djangae/contrib/mappers/pipes.py
+++ b/djangae/contrib/mappers/pipes.py
@@ -1,3 +1,4 @@
+from django.conf import settings
 from mapreduce.mapper_pipeline import MapperPipeline
 from mapreduce import parameters
 from mapreduce import control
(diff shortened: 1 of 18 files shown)


Kazade:
@adamalton This is all yours to review :)

adamalton:
@Kazade I think this is now good to go, with 1 exception... if you run `NamespaceTests` on its own then `test_move_objects_between_namespaces` fails.  I think this is because we're not flushing the namespaces properly between tests, and presumably the tests get run in a different order when you specify that specific test case.

Are you able to take a look and see if you can work out the right way to fix it?

adamalton:
💃🎉

PR #522, opened 2015-12-16
Allow User ID to be blank

jscissr:
For pre-creation.

You still have to fill in a password when pre-creating, I don't know how to fix that.

### DIFF
--- a/djangae/contrib/gauth/common/models.py
+++ b/djangae/contrib/gauth/common/models.py
@@ -41,7 +41,7 @@ class GaeAbstractBaseUser(AbstractBaseUser):
     username = models.CharField(
         # This stores the Google user_id, or custom username for non-Google-based users.
         # We allow it to be null so that Google-based users can be pre-created before they log in.
-        _('User ID'), max_length=21, unique=True, null=True, default=None,
+        _('User ID'), max_length=21, unique=True, null=True, blank=True, default=None,
         validators=[
             validators.RegexValidator(re.compile('^\d{21}$'), _('User Id should be 21 digits.'), 'invalid')
         ]


adamalton:
Thanks!

As for the password, are you creating users via the Django admin?  If so then this might work (in djangae/contrib/gauth/datastore/admin.py):

```python
from django.contrib import admin

# DJANGAE
from djangae.contrib.gauth.datastore.models import (
    GaeDatastoreUser,
    Group
)

class UserAdmin(admin.ModelAdmin):
    exclude = ('password',)

admin.site.register(GaeDatastoreUser, UserAdmin)
admin.site.register(Group)
```

I don't have time to test that right now, but can you try it, and if it works submit a pull request with the same sort of thing added to gauth/sql/admin.py as well?

jscissr:
Um, I noticed that this PR actually doesn't work (it also doesn't harm), as the field will be the empty string instead of None. I wanted to update the PR with a working solution when my debugging is finished, but now I'll just open a second one.

Is there another way to precreate users then, other than the Django admin?

adamalton:
Ah yeah, I knew I would regret merging something without checking it first!  Reverted.

Maybe we need to add `default=None` as well as `blank=True`.  I *think* that will mean that it gets set to `None` when you leave the form field empty (but I'm not 100% sure).  Try it out and open another PR when you've found the solution.

jscissr:
Sorry for that.

My solution now looks like this:

```
class EmptyStringToNoneFormField(forms.CharField):
    def clean(self, value):
        return value or None

class EmptyStringToNoneField(models.CharField):
    def to_python(self, value):
        return value or None
    def formfield(self, **kwargs):
        defaults = {'form_class': EmptyStringToNoneFormField, 'required': False}
        defaults.update(kwargs)
        return super(EmptyStringToNoneField, self).formfield(**defaults)

@python_2_unicode_compatible
class GaeAbstractBaseUser(AbstractBaseUser):
    """ Absract base class for creating a User model which works with the App
    Engine users API. """

    username = EmptyStringToNoneField(
        # This stores the Google user_id, or custom username for non-Google-based users.
        # We allow it to be null so that Google-based users can be pre-created before they log in.
        _('User ID'), max_length=21, unique=True, null=True, default=None,
        validators=[
            validators.RegexValidator(re.compile('^\d{21}$'), _('User Id should be 21 digits.'), 'invalid')
        ]
    )
```

I don't know if there is a better solution, but I tried to find one and didn't.

PR #525, opened 2015-12-17
User pre-creation from Django admin

### LINKED ISSUE #97: Create an interface for pre-creating users
Djangae.contrib.gauth provides the ability to pre-create user objects so that you can create a user before knowing what their Google User ID is (i.e. before they've logged in).  Currently this is only possible programmatically, so it would be good to add an interface for doing this, the Django admin being the most obvious place to do it.

User pre-creation is done by setting the `username` to `None` and only providing the `email`, so we should provide an admin form which allows that.  We should also remove the `password` field from the form, because 1) we don't use passwords for Google-Accounts-based users, and 2) in its raw form that field contains the hash value, so it's not really user-editable anyway.

This custom admin form should be automagically registered with the admin site if the user model in use is `GaeUser` or `GaeDatastoreUser` (see existing code in gauth/admin.py).  But it should also be available for use/subclassing in projects which use a custom user model (i.e. a subclass of `GaeAbstractUser` or `GaeAbstractDatastoreUser`).

jscissr:
Hide password field and make username optional.

See #522. This partially fixes #97.

### DIFF
--- a/djangae/contrib/gauth/datastore/admin.py
+++ b/djangae/contrib/gauth/datastore/admin.py
@@ -1,4 +1,5 @@
 from django.contrib import admin
+from django import forms
 
 # DJANGAE
 from djangae.contrib.gauth.datastore.models import (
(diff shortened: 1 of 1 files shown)


jscissr:
ping @adamalton

adamalton:
I've reverted the 2nd of those commits and removed the form which wasn't actually needed.

adamalton:
Thanks for the PR @jscissr!

PR #533, opened 2015-12-20
Yield pipeline

jacobg:


jacobg:
support for yielding a pipeline

PR #534, opened 2015-12-20
Finalize mapreduce pipeline

jacobg:
Add finalize callback for mapreduce pipeline, because output counters are not ready in regular callback.


Kazade:
Hopefully this (and your other PRs) have been tackled by the helpers in the mapreduce processing module that's now in master. Please let me know if that isn't the case! Thanks!

PR #546, opened 2016-01-06
Fix __year index lookup

olasitarska:
So this pull request is fixing the __year lookup, that was throwing UnSupportedError for a while, and that bug was hidden by our setup to automatically skip Django tests that throw UnSupportedError.

We incorrectly validated query value [here](https://github.com/potatolondon/djangae/commit/632113d03ccd34c05116723fe4f141677f37c725#diff-5f1fd58426c71048b283812e6198689eR122). The value in this case was an integer (year), and `validate_can_be_indexed` method checks for datetime (that's a correct check, but for **indexing**, not lookup). 

I also added a test for `__year`, and fixed another failing test that was hidden because of that bug.


Kazade:
Beautiful!

PR #555, opened 2016-01-13
Json serialize set fields

### LINKED ISSUE #554: When serialized with JSON, SetField instances should be a list, not a string
Djangae monkey-patches DjangoJSONEncoder to handle SetField values, but it uses a string representation. It would be more useful if a SetField value was serialized as a list, same as ListField.

https://github.com/potatolondon/djangae/blob/master/djangae%2Fpatches%2Fjson.py

Would a patch to implement this be accepted?

davidwtbuxton:
This patch changes the serialization behaviour so that SetField instances, and any sets, are encoded as lists. Previously there were encoded as a string literal using the set literal syntax.

This fixes #554 .

### DIFF
--- a/djangae/patches/json.py
+++ b/djangae/patches/json.py
@@ -4,8 +4,8 @@ def additional_type_handler(func):
     @wraps(func)
     def _wrapper(self, o):
         if isinstance(o, set):
-            # Return a string representing a set
-            return "{" + ",".join([str(x) for x in o]) + "}"
+            # Treat sets as lists.
+            return list(o)
         else:
             return func(self, o)
 
--- a/djangae/tests/test_db_fields.py
+++ b/djangae/tests/test_db_fields.py
@@ -1,9 +1,10 @@
+import json
 from collections import OrderedDict
 
 # LIBRARIES
+from django.core import serializers
 from django.db import models
 from django.db.utils import IntegrityError
-from django.contrib.contenttypes.models import ContentType
 
 # DJANGAE
 from djangae.db import transaction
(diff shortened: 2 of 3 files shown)


davidwtbuxton:
The objects in the list will be in an arbitrary order when converted from a set. _Might_ be an idea to use a stable order by calling `sorted(obj)` instead of `list(obj)`, I don't think there would be any unwanted side-effects when comparing objects in the sequence. Then again, it is a set, so pretending there is an order is misleading.

Kazade:
So I was going to write a whole thing about how Django < 1.9 doesn't allow serializing and deserializing of custom fields from anything but strings, and how I didn't want to monkey patch a JSON encoder. Then I saw that someone already had, and then I saw that that someone was me. I'm slightly annoyed at past me.

So given that we're already monkey patching stuff, I'm happy for this to be merged, but, could you also make sure that ListField, RelatedSetField and RelatedListField all, also, serialize to lists? Just because we're introducing an inconsistency at the moment with ListField vs SetField (I think there's already an inconsistency with RelatedX doing the string thing so we should fix that too).

davidwtbuxton:
Would these tests be better in test_serializers or in test_db_fields ?

Kazade:
Probably fields? I guess this is very field specific...

olasitarska:
@davidwtbuxton some tests are failing here -- are you planning to continue work on this?

dannymilsom:
I've only just seen this PR which would also fix same issue described in https://github.com/potatolondon/djangae/issues/881

grzes:
(this code is included and built upon in pull request 902 - look at that instead please)

PR #563, opened 2016-01-26
Allow use of queue_name to run map-reduce tasks on a specific queue

### LINKED ISSUE #562: MapReduceTask always runs on default queue
The [Djangae docs say you can use the queue_name class attribute][docs] to get a map-reduce task to run on queues other than default.

But setting the queue_name attribute has no effect. Tasks always run on the default queue.

[docs]: http://djangae.readthedocs.org/en/latest/mappers/#map-reduce

davidwtbuxton:
Fixes #562


davidwtbuxton:
This was also fixed in #535 , which hasn't been merged.

PR #583, opened 2016-02-15
Add djangae.db.utils.ensure_instance_included

Kazade:
This is a utility function for the common use case of creating an instance and then returning to an inconsistent listing.

Also adds a more obvious error when you try to create an entity with a pk of 0


adamalton:
Would this not make more sense as part of `djangae.contrib.consistency`?  Otherwise we have 2 very similar things in totally different places.

Kazade:
@adamalton I don't think so.. djangae.contrib.consistency is an entire app with a configuration and a bunch of stuff that goes on. This is just a utility function. 

It's similar to `djangae.core.paginator` vs `djangae.contrib.pagination` - lightweight helper vs all-encompassing solution.

adamalton:
I thought that the only reason for `djangae.core.paginator` being in that place was because it matches the location of `django.core.paginator`?

You can use bits of `djangae.contrib.consistency` without installing it as an app - e.g. `improve_queryset_consistency` can be used as a standalone function to filter out recently-deleted objects, even if you don't have any of the configuration stuff set up.

Kazade:
Yeah but that massively confuses things. When you are using stuff from a self-contained Django app you don't know whether or not stuff will work or not if it's not in INSTALLED_APPS. 

I remain unconvinced :)

adamalton:
It's not confusing if it's documented.  But currently there are 2 different things documented in different places with no mention of each other.  That's confusing!

Kazade:
@adamalton er, my documentation specifically mentions djangae.contrib.consistency...

adamalton:
Ah, sorry I didn't see that, I was looking for a link.  It needs a mention/link in the other direction as well then!

Or you could just put these 2 very similar things in the same place :-).

Kazade:
OK after offline discussion we decided to move this to djangae.db.consistency and update the documentation in both djangae.contrib.consistency and in the db backend docs

PR #586, opened 2016-02-15
Make Gauth always add users with their emails lowercased.

### LINKED ISSUE #481: gauth: Lowercase email addresses for lookups and creation
Semi-related to #461.

The fix for #461 removed the `email = email.lower()` line in `common.backends`. Rather than doing that, I think a better fix would have been to add `lower()` to everywhere where the email is normalised (since, `BaseUserManager` doesn't do that as part of normalisation.)

Oftentimes the reason for using user pre-creation is for when someone invites a user to a system and then that user logs in and everything is magically already hooked up.

This doesn't work if user A invites 'person@gmail.com' but user B logs in with 'Person@gmail.com'. User B shouldn't have to care what case their email was written in when they were invited.

Currently, to customise this behaviour, you'd have to make your own copies of `gauth.common.backends.BaseAppEngineUserAPIBackend` and `gauth.middleware.AuthenticationMiddleware`. It'd be nice to either do it by default (but provide a 'migration' to not break everyone else already relying on this behaviour) or have some other way of being able to do it.

Of course ideally the App Engine User's API would just lowercase everything all the time.

olasitarska:
Fixes #481

### DIFF
--- a/djangae/contrib/gauth/common/models.py
+++ b/djangae/contrib/gauth/common/models.py
@@ -32,6 +32,12 @@ class GaeUserManager(UserManager):
         )
         return self.create(**values)
 
+    @classmethod
+    def normalize_email(cls, email):
+        """ Lowercase given email address """
+        email = email or ''
+        return email.lower()
+
 
 @python_2_unicode_compatible
 class GaeAbstractBaseUser(AbstractBaseUser):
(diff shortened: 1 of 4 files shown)


adamalton:
LGTM!

PR #596, opened 2016-02-24
Correcting config in eventual consistency docs 

armirusco:
Currently the docs for the eventual consistency app show an example config that uses camel-case model names. Unfortunately, that seems to fail silently. Converting the model name to lower case does the trick though. 

Updating the docs to make sure it's a bit clearer.

### DIFF
--- a/docs/consistency.md
+++ b/docs/consistency.md
@@ -61,7 +61,7 @@ CONSISTENCY_CONFIG = {
 
     # The settings can be overridden for each individual model
     "models": {
-        "app_name.ModelName": {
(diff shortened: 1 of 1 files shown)


olasitarska:
Just made me wondered if it's an actual bug that we should fix, or just doc typo?

adamalton:
In the Django code, it looks like `_meta.model_name` gets set by `self.model_name = self.object_name.lower()`, and `_meta.model_name` is what we're comparing the value to, so I think that fixing the docs (rather than changing the code) is the right thing to do.

Issue #595, opened 2016-02-23
Simplify settings for auto-creating Django users when using the App Engine authentication backends


davidwtbuxton:
Hello,

The `DJANGAE_FORCE_USER_PRE_CREATION` and `DJANGAE_ALLOW_USER_PRE_CREATION` settings control whether the AppEngineUserAPIBackend allows a _new_ user to authenticate if there is not a corresponding Django user in the database.

It would be nice to simplify the logic for whether a Django user is created and whether the user is allowed access, and to have a more obvious name for the setting(s) that govern this behaviour.

Instead of checking 2 settings, we could have a single setting, something like `DJANGAE_AUTO_CREATE_USERS`. This new setting would default to `False`.

The new logic could be:

- If the user is not logged in with Google Auth then **deny access**.
- If the user is logged in with Google Auth then:
  - If there is no existing Django user and if `AUTO_CREATE_USERS=False` then **deny access**.
  - If there is no existing Django user and if `AUTO_CREATE_USERS=True` then create a new Django user and **allow access**.
  - If there is an existing Django user then update the email address and **allow access**.

One thing missing from the above proposed logic is what should happen if `AUTO_CREATE_USERS=False` and the Google Auth user is an App Engine admin. I'm guessing it would be nice to always create a Django users for App Engine admins and allow access.

It would be nice to respect the existing force/allow settings for backward compatibility, although the existing logic is a little screwy and would be fiddly to maintain exact parity with the proposed new logic (mostly because _if_ there is not an existing Django user **and** `DJANGAE_FORCE_USER_PRE_CREATION=True` **and** the user is not an admin _then_ access is denied).

Also I've no idea if Django then checks the `is_active` flag on an authenticated Django user. Is that something Djangae's authentication backend has to check itself? My assumption is that Django does that after a backend returns a Django user instance.

davidwtbuxton:
Looking at the current implementation, it also does a little dance if there is an existing user with a matching email address but a different user id (stored as the username on the Django user). In that situation it removes the email address from the existing user and then creates a new user with the same username as the existing user (and with the new email address).

I don't see how that can work because GaeAbstractBaseUser.username is unique.

Assuming that logic doesn't work, I think it makes sense to deny access if there is an existing user with a matching email address but a different (non-null) user id.

olasitarska:
has been resolved now! 🎉

PR #600, opened 2016-02-27
Simplify auth backend

davidwtbuxton:
Hello,

These changes replace the 2 settings for the Google Accounts authentication backend with a new setting `DJANGAE_CREATE_UNKNOWN_USER`, as described in #595 .

One difference is that the new logic will not raise `IntegrityError` when `DJANGAE_CREATE_UNKNOWN_USER=False` and there is an existing user with a matching email address.

The setting is named to be consistent with Django's RemoteUser backend, which has a `create_unknown_user` attribute that controls a similar feature to do with auto-creating user accounts.

David B.

### DIFF
--- a/djangae/contrib/gauth/common/backends.py
+++ b/djangae/contrib/gauth/common/backends.py
@@ -21,14 +21,28 @@ if hasattr(settings, "ALLOW_USER_PRE_CREATION"):
     )
 
 
+def should_create_unknown_user():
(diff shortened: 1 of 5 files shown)


olasitarska:
@adamalton would be amazing if you had some time to review this :)

adamalton:
Thanks David.  Much nicer code!

Issue #333, opened 2015-05-14
Improvements to CharField


adamalton:
Things:

1. The Datastore's limit on indexed text fields appears to be a number of **bytes** rather than a number of **characters**.  So on a Django `CharField`, setting `max_length` is not a reliable way to limit the string to the appropriate length for the Datastore.
2. The Datastore's indexable text limit recently went from 500 bytes to 1500 bytes.

Things we should do about those things:

* Create a custom `MaxBytes` validator, which limits on bytes rather than characters.  I.e. `len(value.encode('utf-8'))`, not `len(value)`.  It might be nice to subclass `django.core.validators.BaseValidator` for that.
* Create a custom `CharField` which:
  * Enforces a default `MaxBytes` validator with its `limit_value` set to `google.appengine.api.datastore_types._MAX_STRING_LENGTH`.
  * By default sets `max_length` to `google.appengine.api.datastore_types._MAX_STRING_LENGTH`.  This means that `max_length` becomes an optional kwarg, but is still overridable if you want to limit the length to something less than App Engine's limit.

owad:
#604

PR #604, opened 2016-03-03
Improvements to CharField (issue #333)

owad:
- add MaxBytesValidator class
- add custom djangae CharField with max_lenght set to 1500 bytes (current GAE
  limit) by default and MaxBytesValidator added to its validators
- add tests for the new field


Kazade:
Outstanding! Thank you!

PR #608, opened 2016-03-08
1 9 support

owad:


Kazade:
I've cherry picked your ordering test fix, but the special indexing stuff needs more thought. Thanks though!

PR #617, opened 2016-03-11
Adds ensure_instances_consistent, and deprecates ensure_instance_included

Kazade:
Better naming, more features, such wow.


asendecka:
Looks good! Great job :+1:

adamalton:
There's now an unnecessary try/except in here.  Fixing...

adamalton:
See #620.

PR #620, opened 2016-03-14
Removing unnecessary try/except from db.consistency

adamalton:
There used to be a `.get()` in here, but now it's a `.filter()` so the try/except is unnecessary.


asendecka:
Good spot! Thanks :star:

PR #623, opened 2016-03-16
Documentation updates/tweaks.

adamalton:
* Updating statement about supported Django versions.
* Adding documentation for computed fields.
* Clarifying that the `CONSISTENCY_CONFIG` belongs in settings.py.


olasitarska:
Nice one, thanks!

Issue #624, opened 2016-03-16
Gauth backend raises IntegrityError when creating a new user


davidwtbuxton:
A user is reporting errors when using Google Auth and Djangae's gauth backend, but only some of the time. Looks like it might be a race condition with 2 requests coming in very close together.

In their settings file, `DJANGAE_ALLOW_USER_PRE_CREATION=True` and other auth-related settings are not set.

Using Djangae commit 5b9bdd77d012293519d5df23c7d2022b39c44ac3

```
Internal Server Error: /
Traceback (most recent call last):
  File "/base/data/home/apps/s~myproj.12345/third_party/django/core/handlers/base.py", line 108, in get_response
    response = middleware_method(request)
  File "/base/data/home/apps/s~myproj.12345/third_party/djangae/contrib/gauth/middleware.py", line 25, in process_request
    django_user = authenticate(google_user=google_user) or AnonymousUser()
  File "/base/data/home/apps/s~myproj.12345/third_party/django/contrib/auth/__init__.py", line 74, in authenticate
    user = backend.authenticate(**credentials)
  File "/base/data/home/apps/s~myproj.12345/third_party/djangae/contrib/gauth/common/backends.py", line 69, in authenticate
    return User.objects.create_user(user_id, email)
  File "/base/data/home/apps/s~myproj.12345/third_party/django/contrib/auth/models.py", line 187, in create_user
    **extra_fields)
  File "/base/data/home/apps/s~myproj.12345/third_party/django/contrib/auth/models.py", line 182, in _create_user
    user.save(using=self._db)
  File "/base/data/home/apps/s~myproj.12345/third_party/django/db/models/base.py", line 734, in save
    force_update=force_update, update_fields=update_fields)
  File "/base/data/home/apps/s~myproj.12345/third_party/django/db/models/base.py", line 762, in save_base
    updated = self._save_table(raw, cls, force_insert, force_update, using, update_fields)
  File "/base/data/home/apps/s~myproj.12345/third_party/django/db/models/base.py", line 846, in _save_table
    result = self._do_insert(cls._base_manager, using, fields, update_pk, raw)
  File "/base/data/home/apps/s~myproj.12345/third_party/django/db/models/base.py", line 885, in _do_insert
    using=using, raw=raw)
  File "/base/data/home/apps/s~myproj.12345/third_party/django/db/models/manager.py", line 127, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
[...]

PR #625, opened 2016-03-16
Alias Django's db exceptions to djangae.db.backends.appengine.dbapi.

davidwtbuxton:
Djangae raises its own IntegrityError in at least one place, which is
not the same as Django's IntegrityError, so this change makes them
the same error class.

This does change the exception hierarchy of some errors. Specifically
NotSupportedError, DatabaseError and InterfaceError now have different
super-classes.

### DIFF
--- a/djangae/db/backends/appengine/dbapi.py
+++ b/djangae/db/backends/appengine/dbapi.py
@@ -1,34 +1,24 @@
 """ Fake DB API 2.0 for App engine """
+from django.db import utils
 
-class DatabaseError(Exception):
-    pass
 
-class IntegrityError(DatabaseError):
-    pass
+DatabaseError = utils.DatabaseError
+DataError = utils.DataError
+IntegrityError = utils.IntegrityError
+InterfaceError = utils.InterfaceError
+InternalError = utils.InternalError
+NotSupportedError = utils.NotSupportedError
+OperationalError = utils.OperationalError
+ProgrammingError = utils.ProgrammingError
 
-class NotSupportedError(Exception):
-    pass
 
 class CouldBeSupportedError(NotSupportedError):
     pass
 
-class DataError(DatabaseError):
-    pass
-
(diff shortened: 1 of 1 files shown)


davidwtbuxton:
At the moment Djangae throws its own IntegrityError in djangae/db/constraints.py, which is a pain if we are going to fix #624 because it looks like we will need to catch IntegrityError there, and would need to catch Djangae's own flavour of IntegrityError.

Really Djangae should be using Django's IntegrityError, but this patch is for compatibility so that we don't break any app which happens to import Djangae's versions of these exceptions.

A long-term fix would be to remove most if not all of these exceptions from djangae/db/backends/appengine/dbapi.py.

In commit d34ff6c25d07a7c32fb69bdb468fe4baabc7c8e6 there is this note:

> specifically don't subclass Django exceptions because that just breaks everything

But the tests pass! So it must be safe ;)

adamalton:
Thoughts:

* Is there a need to do `ExceptionName = utils.ExceptionName`?  Surely just importing them would make them available in the same way?
* I was thinking that for things like #624 where we want to catch a specific unique constraint violation that it would actually be good to have a specific `UniqueConstraintViolation` exception (which is a subclass of `IntegrityError`, which could be raised with a dictionary of `{field_name: value}`, thus allowing us to catch it for a specific field.  And because it would be a subclass of `IntegrityError` it would still get handled in the normal way by other bits of code which are doing `except IntegrityError`.

adamalton:
Apparently backends must throw their own exceptions and Django translates them.

So even though Djangae raises its own exception class (as it should), that's not what the exception is when it gets to you (so we can just catch Django's `IntegrityError`).


You can see this in the other backends:  
https://github.com/django/django/blob/master/django/db/backends/postgresql/base.py#L47-L48  
https://github.com/django/django/blob/master/django/db/backends/sqlite3/base.py#L49-L50

Closing this off...

PR #626, opened 2016-03-17
Making auth backend catch race condition when creating a new user.

adamalton:
This is (probably) a fix for #624.

But note that in #624 the `IntegrityError` being raised is for the `email` field, whereas if you undo this fix and run the new test without it then the `IntegrityError` you get is for the `username` field. But also note that the traceback in #624 shows a custom user model.

Also note that this fix catches `django.db.utils.IntegrityError`, whereas if you look at the code where the exception is raised then you would think that it should be catching `djangae.db.backends.appengine.dbapi.IntegrityError`.  I have no idea how that exception morphs itself from one to the other!  Maybe #625 will fix that?!

### DIFF
--- a/djangae/contrib/gauth/common/backends.py
+++ b/djangae/contrib/gauth/common/backends.py
@@ -1,5 +1,6 @@
 import warnings
 
+from django.db.utils import IntegrityError
 from django.db import transaction
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


adamalton:
Kazade has pointed out that backends must throw their own exceptions and Django translates them, which explains the exception mis-match.

PR #536, opened 2015-12-21
Pipelines & MapReduce

chargrizzle:
Adds native pipeline and mapreduce support! 

This will deprecate djangae.contrib.mappers

PR #628, opened 2016-03-18
Ensure MapReduceTask uses the db returned by the application router(s)

sleepyjames:
This PR changes to the default db to be the one returned by the applications router.

### DIFF
--- a/djangae/contrib/mappers/pipes.py
+++ b/djangae/contrib/mappers/pipes.py
@@ -1,4 +1,5 @@
 from django.conf import settings
+from django.db import router
 from mapreduce.mapper_pipeline import MapperPipeline
 from mapreduce import parameters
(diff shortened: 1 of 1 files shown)


Kazade:
Rather than using `model.objects.all().db`, this should instead use `django.db.router.db_for_read(model)` then I'll happily merge this :)

sleepyjames:
Done, however I'm sure this will be superseded by all the work in https://github.com/potatolondon/djangae/pull/536 perhaps ?

Kazade:
@sleepyjames No that's an additional thing, we still need to support the mappers app even if we deprecate it. One of the tests is now failing :(

sleepyjames:
Ok fixed properly this time. Didn't run the uniquetool tests :-|

PR #640, opened 2016-03-29
the --oauth2 option is no longer needed

MihailRussu:
it's now default and used whether 2fa is enabled or not


adamalton:
Thanks!

PR #645, opened 2016-04-02
RelatedIterator fix

kockicica:
Fails when related iterated fields model is set as string


adamalton:
Hi @kockicica, thanks for contributing.

This change appears to break one of the existing tests though, see the [test report](https://travis-ci.org/potatolondon/djangae/jobs/120264564#1-5467).  Could you update your fix to avoid that issue so that we can then merge it?

Thanks

Kazade:
Thanks!

PR #647, opened 2016-04-13
Update runserver.py

owad:
Set module_settings to the default value (DJANGO_SETTINGS_MODULE) if module_name is 'default' OR None.


olasitarska:
excellent!

PR #648, opened 2016-04-15
Adds ability to override the application to run remote commands against.

allofthesepeople:
Additional arg `--application` to override the application used in app.yaml, or where app.yaml does not specify an application.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,8 @@
 
 ### New features & improvements:
 
-- Cleaned up and refactored internal implementation of SimulatedContentTypeManager. Now also allows patching ContentType manager in migrations. 
--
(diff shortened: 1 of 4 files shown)


olasitarska:
I'm wondering whether that should be `--app_id` instead of `--application` to keep consistency with devapp server? 

Also, this pull request needs rebasing from master and changelog.md entry.

Thanks!

allofthesepeople:
Make’s sense: had originally kept the naming convention in line with `app.yaml` 

Will update over the next few days

olasitarska:
Thanks!

allofthesepeople:
@olasitarska updated, albeit much later than expected. apologies

olasitarska:
Thank you!

PR #649, opened 2016-04-15
Allow the Mapreduce status UI to function when a CSP is enforced with django-csp.

sleepyjames:


adamalton:
Discussed this, and decided that @sleepyjames will see if we can use the hash method for allowing these specific scripts, rather than just allowing anything on these pages, assuming that's not too much of a pain to do.

olasitarska:
@sleepyjames any updates here?

sleepyjames:
Hey. Sorry the project I was using this on got paused. So I tried the hash approach but it doesn't work for me as the inline JS changes. I think the only way will be to apply the unsafe-inline policy to only the views that require it.

olasitarska:
Let's go with this for now! Thanks @sleepyjames :)

PR #651, opened 2016-04-20
Do not set all GAE users as superusers if hardcoded list of superuser…

owad:
…s provided.


adamalton:
Rather than do this, I've created another PR (#691) which does this slightly differently and more generically.

Having a list of admins set in the code seems like quite a project-specific thing, and some projects might want to do it differently (e.g. have a list of superusers in the DB, or be able to set the superusers manually).

So my new PR just lets you decide whether or how the GAE admins are set as superusers or not.  And then if you want to have a list of superusers in the code, you could implement that by just listening to the `django.contrib.auth.signals.user_logged_in` signal and modifying the user in there.

Gonna close this.

PR #652, opened 2016-04-25
Fix JSONField behaviour in forms

adamalton:
Previously, if we had a JSONField on a model, and we edited that in a form, then we would take the raw string from the POST and essentially pass it straight through to the field in the DB.  If you then reloaded the object from the DB then it would parse the JSON back into python, but there were 2 problems:

1. We didn't do any validation on the JSON string before saving it (i.e. didn't attempt to parse it).
1. If you overrode your form's `clean` method and accessed `self.cleaned_data['my_json_field']` then you would have a string, not the proper object (which is contrary to how things like BooleanFields work, where `cleaned_data` gives `True` or `False`, not `"1"` or `""`.

This pull request fixes that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,12 @@
 
 ### Bug fixes:
 
--
+- Fix JSONField behaviour in forms: it's properly validating JSON string before saving
+it and returns json object, not string when accessed through cleaned_data. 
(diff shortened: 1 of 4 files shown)


olasitarska:
Thanks @adamalton!

PR #658, opened 2016-04-28
Add custom test assertions to check if login:admin or login:required protect a handler

dannymilsom:
When writing tests for an app it would be useful to have some built in assertion methods to verify that handlers defined in app.yaml have the correct login configuration.

This would reduce the likelihood of that line being removed and the change not being noticed (it could have profound security implications after all).

It would also make it more explicit that requests processed by a handler require users to login in the tests. Currently this can be quite transparent.

Obviously these assertions don't need to test that the configuration is correctly processed by appengine (those tests already exist in the SDK), but only assert that presence of the correct configuration for a defined handler.


adamalton:
Thanks @dannymilsom !

PR #659, opened 2016-05-06
BlobstoreFileUploadHandler supports brackets in file input name

jwicks:
For example, support `<input type="file" name="files[]">`

### DIFF
--- a/djangae/storage.py
+++ b/djangae/storage.py
@@ -423,7 +423,7 @@ class BlobstoreFileUploadHandler(FileUploadHandler):
                 continue
 
             # OK, we have a blob key, but is it the one for the field?
-            match = re.search('\sname="?(?P<field_name>[a-zA-Z0-9_-]+)', part)
+            match = re.search('\sname="?(?P<field_name>[a-zA-Z0-9_\-\[\]]+)', part)
             name = match.groupdict().get('field_name') if match else None
             if name != field_name:
                 # Nope, not for this field
--- a/djangae/storage.py
+++ b/djangae/storage.py
@@ -423,7 +423,7 @@ class BlobstoreFileUploadHandler(FileUploadHandler):
                 continue
 
             # OK, we have a blob key, but is it the one for the field?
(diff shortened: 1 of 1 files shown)


adamalton:
In thinking that that character class in the regex should just be [^"].
I.e. anything which isn't a closing quotation mark.

On Friday, 6 May 2016, Jon Wicks <notifications@github.com> wrote:

> For example, support <input type="file" name="files[]">
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/potatolondon/djangae/pull/659
> Commit Summary
>
>    - BlobstoreFileUploadHandler supports brackets in file input name
>
> File Changes
>
>    - *M* djangae/storage.py
>    <https://github.com/potatolondon/djangae/pull/659/files#diff-0> (2)
>
> Patch Links:
>
>    - https://github.com/potatolondon/djangae/pull/659.patch
>    - https://github.com/potatolondon/djangae/pull/659.diff
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/potatolondon/djangae/pull/659>
>


-- 
Sent from my phone; please ignite autocorrupt errors

jwicks:
Updated. I went with `[^\'"]`

adamalton:
Thanks Jon,

However... I don't think we should allow a single quote to _close_ the string if we don't allow single quotes to open it.  I.e. we do `name="` (double quote) so I think we should only allow double quotes to close it.  Otherwise people doing things like `name="Jon's File" will be in trouble!

We _could_ allow single quotes at the start (`name='`) but then we'd have to only allow a single quote at the end if it was a single quote at the start, and only allow double quotes at the end if it was a double quote at the start.  And then it all descends into a painful world of regex complexity.

So I think we should just go for ` [^"]` and leave at that unless someone complains!  :-)

jwicks:
Ah, sounds good. I overlooked the opening quote. Updated to `[^"]`

adamalton:
Great.  Thanks Jon!

PR #668, opened 2016-05-17
Use a stack for state in the atomic + non_atomic decorators. Fixes #667

### LINKED ISSUE #667: Recursive use of atomic() or non_atomic() decorators is broken
The problem is that atomic and non_atomic have a variable called `state`. This is thread local, but if you recursively call a decorated function the state gets overwritten which introduces subtle bugs!

Kazade:


adamalton:
Needs comments!  😛

PR #672, opened 2016-05-24
Make sure libraries in sitepackages directory have precedent over bundled appengine libs

dannymilsom:
- Previously if two libraries were in both sitepackages and appengine/libs, the later would be resolved.
- Now any locally managed third party libraries take precedent.

This is a workaround for the bigger problem of all appengine/lib paths being added to sys.path. Only the libraries specified in app.yaml should be available for use (this currently leads to a difference in local and production behaviour).


adamalton:
As per the internal discussion, this still leaves some of the GAE libs at the beginning of `sys.path` (antlr3, yaml, webapp2, webob, etc), which is presumably because dev_appserver shoehorns them in there.  This is still an improvement on what we had before though (i.e. this means that fewer of the SDK libs come before the local ones!).

PR #673, opened 2016-05-25
Fix iexact index with underscores

adamalton:
Fixes issue described on mailing list: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/djangae-users/erIUL3LZ9GA/ZWo0OAl1AQAJ

`__iexact` queries now work correctly when the indexed value contains an underscore.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### New features & improvements:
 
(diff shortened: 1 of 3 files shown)


olasitarska:
Yay, thanks!

PR #674, opened 2016-05-25
Work in progress: Django 1.10 support

olasitarska:
This is to track the work being done. Not ready to merge until stuff passes :)


djangae/patches/contenttypes.py
```diff
@@ -139,6 +140,8 @@ def disable_save(*args, **kwargs):
         if dic["id"] in self._store.constructed_instances:
             return self._store.constructed_instances[dic["id"]]
         else:
+            if django.VERSION[1] > 9:
```
timgraham: I think this condition (with slicing) is more cryptic than something like `django.VERSION > (1, 9)`

Also, if I were doing this, I would try to make it so that the default path works for the newest version of Django so that when dropping old versions, you can simply remove the entire if statement instead of figuring out what else has to change. (i.e. only add the 'name' key in the dictionary on line 70 for older versions).

.travis.yml
```diff
@@ -5,6 +5,7 @@ python:
 env:
   - "DJANGO_VERSION=1.8"
   - "DJANGO_VERSION=1.9"
+  - "DJANGO_VERSION=master"
```
timgraham: You'll need to test against Django 1.10 alpha or the stable/1.10.x branch as master is now 1.11 pre-alpha.

olasitarska:
@timgraham Both good suggestions! I fixed that now, but we just hit a new issue with 1.10 support: #682

adamalton:
This now just needs some documentation adding for the new list/set field filtering.

PR #679, opened 2016-05-27
Fixes #630 - Add Django master to Travis builds we run, but allow it to fail

### LINKED ISSUE #630: Run Django master test via Travis daily
We wanna setup Travis in a way that it runs Djangae master against Django master tests once a day, and reports to our Slack so we can keep an eye on future Django release

olasitarska:
I think this is gonna satisfy our requirements of having all of our tests be run on Django master, but without interrupting current development flow. This config runs of Django 1.8, 1.9 and master, but allows tests run on master to fail while still being marked as green. 

Nightly builds are already configured and happening.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,17 @@
 language: python
 sudo: false
 python:
-- '2.7'
+- 2.7
 env:
-  - "DJANGO_VERSION=1.8"
-  - "DJANGO_VERSION=1.9"
-
+  - DJANGO_VERSION=1.8
+  - DJANGO_VERSION=1.9
+  - DJANGO_VERSION=master
+matrix:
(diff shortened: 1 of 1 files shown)


timgraham:
Maybe you can modify the `djangae/db/backends/appengine/query.py` factories to use a default one if the VERSION key isn't present (or otherwise add entries for (1, 10) and (1,11) so the build at least gets a bit farther than those failures)?

olasitarska:
This PR is now suppressed by #683, so closing.

@timgraham thanks for that suggestion, added it in #683

timgraham:
Great. I think "superseded" is the word you're looking for there. :-)

olasitarska:
Duh, yes :)

PR #680, opened 2016-05-31
Make sure the ordering of paths on sys.path is correct

Kazade:
Fixes bug in master.

- Fixes a bug introduced in a previous PR which caused system-wide paths to have priority over SDK ones

PR checklist:
- [x] Updated CHANGELOG.md (existing changelog entry is still OK)
- [x] Added tests for my change (sandbox changes, so no test necessary/possible)

The previous change broke things by prioritising system-wide python paths over
App Engine ones. This fixes that by only prioritising paths which are under the project
root and making sure all other paths come after the SDK.

### DIFF
--- a/djangae/sandbox.py
+++ b/djangae/sandbox.py
@@ -8,6 +8,8 @@ import getpass
 import logging
 import urllib
 
+from os.path import commonprefix
+
 from . import environment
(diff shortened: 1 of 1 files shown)


olasitarska:
Very pretty!

PR #683, opened 2016-06-01
Fixes to make Djangae run on Django 1.10 alpha 

olasitarska:
So these are a couple of fixes that we can merge now to master even though tests don't pass on Django 1.10 alpha 1 and master branches yet. They're marked as allowed to fail in Travis setup though. 

Fixed in this PR:
- Adds 1.10a1 and master to Travis config. Allows them to run and fail. Changes to install_deps and settings to allow for version number 1.10a1
- We were calling `super` on RelatedSetField in RelatedListField
- Model reference in ShardedCounterField needs to be a string, otherwise Django migrations fail on creating testing db
- Added more flexibility so 1.10 or 1.11 doesn't fail because they're not in query factories.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,17 @@
 language: python
 sudo: false
 python:
-- '2.7'
+- 2.7
 env:
-  - "DJANGO_VERSION=1.8"
-  - "DJANGO_VERSION=1.9"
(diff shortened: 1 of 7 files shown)


olasitarska:
Probably none of this needs a changelog, as these are internal only changes?

PR #684, opened 2016-06-02
Allow precreated users to login when DJANGAE_CREATE_UNKNOWN_USER is False

moggers87:
- Added a test for pre-created users when DJANGAE_CREATE_UNKNOWN_USER is False
- Moved auto_create/admin check to allow pre-created users to be fetched as they would have been pre-0.9.4

PR checklist:
- ~~Updated relevant documentation~~ N/A
- [x] Updated CHANGELOG.md 


adamalton:
Thanks!

PR #685, opened 2016-06-03
Fixes #682 - Move ContentType patch to djangae.contrib.contenttypes

### LINKED ISSUE #682: Move the simulated contenttypes into djangae.contrib.contenttypes
In djangae/apps.py we patch contenttypes to use "simulated contenttypes" which are an in-memory approximation of the normal content type behaviour. This is necessary for a couple of reasons:

1. `django.contrib.contenttypes` uses bulk_create, which with the appengine backend is limited to 25 entities - this causes migrate to explode
2. `django.contrib.contenttypes` suffers from eventual consistency issues on the datastore

We have to monkey-patch things because the `ContentType` model is not swappable, however, we need to patch things after contenttypes has loaded, which means djangae would need to come after `django.contrib.contenttypes` in `INSTALLED_APPS` which isn't ideal.

Instead it will be far cleaner to add `djangae.contrib.contenttypes` which must be in `INSTALLED_APPS` after `django.contrib.contenttypes`. The mere inclusion of this app will do the patching removing the need for the `DJANGAE_SIMULATE_CONTENTTYPES` setting.

Going forward it might be worth seeing if we can send a patch upstream to allow making the ContentType model swappable so we can provide a much cleaner solution to this.

olasitarska:
Fixes #682 

- Moves ContentType patch to djangae.contrib.contenttypes
- If django.contrib.contenttypes is in INSTALLED_APPS, but not djangae.contrib.contenttypes, raises an ImproperlyConfigured error

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Extended runserver's file watcher patching to allow ignoring of directories.
 - Add tasks utility functions to djangae.environment.
 - Alias DatastorePaginator -> Paginator, and DatastorePage -> Page to be more like Django
+- Moved `ContentType` patching to `djangae.contrib.contenttypes`. `DJANGAE_SIMULATE_CONTENTTYPES` setting has been deprecated, add `djangae.contrib.contenttypes` to `INSTALLED_APPS` instead. `djangae.contrib.contenttypes` needs to be after `django.contrib.contenttypes` in the `INSTALLED_APPS` order.
 
 ### Bug fixes:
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - Extended runserver's file watcher patching to allow ignoring of directories.
 - Add tasks utility functions to djangae.environment.
 - Alias DatastorePaginator -> Paginator, and DatastorePage -> Page to be more like Django
+- Moved `ContentType` patching to `djangae.contrib.contenttypes`. `DJANGAE_SIMULATE_CONTENTTYPES` setting has been deprecated, add `djangae.contrib.contenttypes` to `INSTALLED_APPS` instead. `djangae.contrib.contenttypes` needs to be after `django.contrib.contenttypes` in the `INSTALLED_APPS` order.
```
timgraham: Looks like the setting was removed rather than deprecated?

PR #686, opened 2016-06-07
Move changelog entry for #684 from 0.9.5 to 0.9.6.

adamalton:
#684 was committed before the 0.9.5 release but was not merged until afterwards, and hence the CHANGELOG entry ended up in the wrong section.  This fixes that.


moggers87:
LGTM

Issue #342, opened 2015-05-19
Allow args to set local service account to be passed to runserver command


martinallison:
Passing

```
--appidentity_email_address APPIDENTITY_EMAIL_ADDRESS
--appidentity_private_key_path APPIDENTITY_PRIVATE_KEY_PATH
```

to `dev_appserver.py` allows you to set the service account for local dev, it would be good if we could pass them to `runserver` somehow.

martinallison:
Adding these to the `runserver` command itself, then setting them on `sandbox._OPTIONS` doesn't work because the sandbox has already been setup by then. Passing them as overrides to `exec_from_command_line` does work, though, because the sandbox is initialised before `runserver` actually happens.

nealedj:
#688 should address this.

grzes:
@martinallison After the 688 merge code seems to pass those vars correctly, but I'm not sure where to get the private key/email pair to actually verify this. It would be awesome if you could check that the latest master does what you need when you have a minute.

PR #688, opened 2016-06-08
Pass remainder of args to the dev_appserver arg parser.

### LINKED ISSUE #614: Passing --require_indexes=yes doesn't set the option on dev_appserver
Run `manage.py runserver --require_indexes=yes`.

`datastore_require_indexes` (devappsever2.py:901) is set to `False`

nealedj:
Fixes #614 .

It appears that none of the `WHITELISTED_DEV_APPSERVER_OPTIONS` defined in `runserver` actually get passed to the sandbox server. The arg parser doesn't read any args from the command line as we're explicitly passing the args but missing the command args.


nealedj:
Fixes #342

adamalton:
I'm going to leave @Kazade to review this, because he understands the sandbox stuff way better than me.  But in the meantime... it needs a CHANGELOG entry :-)

nealedj:
Ah yeah - done!

Kazade:
@nealedj sorry could you bring this up to date for me?

olasitarska:
Bouncing this back to @nealedj -- this fixes it, but it's probably not the best fix, as we may fall into a problem when we need to pass more arguments to `runserver`, and we also don't wanna pass _anything_ to `devappserver`, only arguments that are on whitelisted list. There is functionality for it with `WHITELISTED_DEV_APPSERVER_OPTIONS`, but apparently broken!

nealedj:
Have re-done this to strip the args down to only those in `WHITELISTED_DEV_APPSERVER_OPTIONS`

PR #691, opened 2016-06-14
Gauth supuser sync mode

adamalton:
Currently, the gauth middleware sets `is_staff` and `is_superuser` on all users based on whether or not they are admins of the App Engine application.  This is generally useful but sometimes annoying, as it means that (1) you can't have any superusers which are not GAE admins (because gauth sets their superuser flag back to `False)`, and (2) it forcefully makes all GAE admins into superusers, which is not necessarily what you want, especially if you have some which only have the "viewer" role.

This pull request replaces that forceful, one-size-fits-all approach with a setting to allow you to choose whether or how the users' superuser status is sync'd with their GAE admin status.  So you can now choose to have

* Enforced syncing (like the current behaviour).
* Semi-syncing (GAE admins are set to superuser, but you can set other users to be superusers too, and they don't get altered).
* No syncing at all.

The mode is controlled via the new `DJANGAE_SUPERUSER_SYNC_MODE` setting.

PR checklist:
- [x] Updated CHANGELOG.md 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### New features & improvements:
 
--
+- Added a `DJANGAE_SUPERUSER_SYNC_MODE` setting to allow control of whether and how the gauth middleware updates users' `is_superuser` based on whether or not they are an admin of the App Engine application.
 
 ### Bug fixes:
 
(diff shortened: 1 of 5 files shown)


Kazade:
After discussion we've decided a better way to deal with this is to break up the existing middleware so that the "sync" part can be overridden.

adamalton:
This was fixed by #711

PR #696, opened 2016-06-17
Make testapp runserver work

adamalton:
- Configuration changes to allow the `runserver` command to work in the `testapp` project that's bundled in the repo.

The `testapp` is generally only used for running tests, and so these changes aren't essential.  But it kind of sucks if you want to test something with `runserver` only to find that you have to use djangae-scaffold for it because the `testapp` isn't configured properly.

PR checklist:
- [ ] Updated relevant documentation - N/A
- [x] Updated CHANGELOG.md 
- [ ] Added tests for my change = N/A


adamalton:
💃

PR #697, opened 2016-06-20
Fix `ValidationError` import error in JSON field.

alfredo:
The `ValidationError` exception is not imported in the file, it can be reached from `forms`.

- Fixes import error.


adamalton:
Thanks!

PR #709, opened 2016-07-07
Change `test_ordering_on_sparse_field` to avoid intermittent failure.

adamalton:
`test_ordering_on_sparse_field` tests the order in which objects are returned when the field that is being ordered on is missing.  It correctly checks that the objects with a missing value are returned before objects *with* a value, but it also expects the objects with a missing value to come back in a specific order.

This change removes the ambiguity of the ordering of items with a `color` value of `None`.

PR checklist:
- [x] Updated CHANGELOG.md 


adamalton:
\o/

PR #712, opened 2016-07-08
Change the default for ALLOWED_HOSTS to '*'

### LINKED ISSUE #421: Suggestion: remove ALLOWED_HOSTS restriction
Given that App Engine & Google Apps manages the mapping of a domain name to an App Engine application and defines which host an App Engine app can respond to ALLOWED_HOSTS is redundant.

Perhaps this restriction should be removed to aid deployment of apps and confusion when they return 400s in different environments?

Kazade:
Fixes #421 

- Change the ALLOWED_HOSTS default

PR checklist:
- [X] Updated CHANGELOG.md 

Google takes care of defending against host header attacks and having this fixed at
any other value has caused confusion when deployed apps return 400 responses

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ### New features & improvements:
 
--
+- ALLOWED_HOSTS is now set to ("*",) by default as App Engine deals with routing and this prevents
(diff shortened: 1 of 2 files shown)


adamalton:
Where's the CHANGLOG entry?

PR #720, opened 2016-07-25
Remove `count()` method from `RelatedShardManager`.

### LINKED ISSUE #719: ShardedCounterField should disable .count()
I've lost count of the number of times I've made this dangerous mistake:

```
count_of_things = instance.my_counter.count()
```
This would return the number of shards, not the value of the counter - which is returned by `.value()`

We should disable `.count()` and replace it with `shard_count()`

adamalton:
Fixes #719.

- Remove ambiguous `count()` method from `RelatedShardManager`.
- Add `shard_count()` method to take its place.

PR checklist:
- [x] Updated CHANGELOG.md 

The function of `count()` was ambiguous/misleading.  Did it mean the count of the number of shards, or the sum total of the shard values?

This replaces it with a `shard_count()` method.


olasitarska:
Not sure if we wanna do any deprecation here, as it will break things now for people using `.count()` intentionally?

olasitarska:
Other than that, PR looks good, so I'm happy to merge this as is!

adamalton:
Ok, changed it to a deprecation.

PR #722, opened 2016-07-27
Fix user email case sensitivity once and for all

adamalton:
Fixes #everything.  This will definitely not come back and bite me again.

Hopefully this removes all issues around the uppercase vs. lowercase differences in email addresses in Gauth.  So now:

* Neither the case of the email address of pre-created `User` object nor the case of the email address of the Google that a pre-created user logs in with make any difference.
* Email addresses are stored case sensitively, so that sending emails to people always works.
* Email addresses are forced to be unique on a case-insensitive basis, so:
  - You cannot create _a@example.com_ and _A@example.com.
  - In the rare case where the email address on a Google account changes and the original email address moves to a different Google account, it can change case when it moves and we will still handle it.
* All is good in the world.

PR checklist:
- [x] Updated CHANGELOG.md 

### DIFF
--- a/djangae/contrib/gauth/common/models.py
+++ b/djangae/contrib/gauth/common/models.py
@@ -28,18 +28,13 @@ class GaeUserManager(UserManager):
         values.update(**extra_fields)
         values.update(
             # things which cannot be overridden
-            email=self.normalize_email(email),
+            email=self.normalize_email(email),  # lowercase the domain only
             username=None,
             password=make_password(None),  # unusable password
             # Stupidly, last_login is not nullable, so we can't set it to None.
         )
         return self.create(**values)
 
-    @classmethod
-    def normalize_email(cls, email):
-        """ Lowercase given email address """
-        email = email or ''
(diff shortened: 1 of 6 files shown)


olasitarska:
Tried hard to poke some holes in this, but looks solid. Merging! 🌈

adamalton:
\0/ 🍍

djangae/contrib/gauth/common/models.py
```diff
@@ -106,3 +121,33 @@ def __str__(self):
...
+
+        try:
+            super(GaeAbstractBaseUser, self).validate_unique(exclude=exclude)
+        except ValidationError as super_error:
+            pass
+        else:
+            super_error = None
+
+        if self.email and "email" not in exclude:
+            existing = self.__class__.objects.filter(email_lower=self.email.lower()).exists()
```
asendecka: This should exclude `self` - otherwise we are unable to resave the instance, since it will always be existing.

PR #725, opened 2016-07-27
723 on change sharded counter

### LINKED ISSUE #723: on_change callback in ShardedCounter
There are times when we would like to do something with the instance after the `ShardedCounter` was changed (i.e. if we would like to store denormalized counter on our model). 

We would like to pass the function in `on_change` argument and call it with model instance as a parameter.

asendecka:
Fixes #723.

- allow passing `on_change` callback function to `ShardedCounterField`
- call callback function when counter was changed (on `increment`, `decrement` and `reset`)

PR checklist:
- [x] Updated CHANGELOG.md 


adamalton:
Thanks @asendecka! 🌻

PR #726, opened 2016-07-27
Fix bug where existing User objects wouldn't validate due to email unique constraint not excluding self.

adamalton:
Fixes the second thing that I broke in #722.

- See title.

PR checklist:  
(I'm not adding this to the changelog and am going to pretend that this was all done as part of #722.)

- [ ] Updated CHANGELOG.md 


asendecka:
Thanks @adamalton ✨

PR #731, opened 2016-08-01
Fix incorrect console URL stored in Mapreduce status entity.

sleepyjames:
In the call to `set_status` after kicking off a Mapreduce we use the original
BASE_PATH instead of our own custom one when setting the console URL
for the pipeline. This breaks the pipelines status UI as it uses this
value to link to console URL of child pipelines.


olasitarska:
Thanks @sleepyjames! Could you update CHANGELOG.md please?

sleepyjames:
Sure, changelog updated.

olasitarska:
Thanks! 🌈

PR #735, opened 2016-08-03
Clearsessions

### LINKED ISSUE #549: Add a view to delete old sessions
As per the [Django docs](https://docs.djangoproject.com/en/dev/topics/http/sessions/#clearing-the-session-store) old session data should be cleared out regularly.

We should provide a view in `djangae.views` which does this.  Then in [djangae-scaffold](https://github.com/potatolondon/djangae-scaffold) we should add a cron job that calls this view.

chargrizzle:
Fixes #549

-  clearsessions
-  sandbox and patch fixes

PR checklist:
- [x] Updated CHANGELOG.md 


chargrizzle:
Rebased

PR #739, opened 2016-08-16
Repopulate contenttypes on .all() call. Add missing import.

olasitarska:
Some issues with simulated content types I found while testing Wagtail on Djangae. I don't think this needs a changelog/docs change, as this is just some internal implementation issues?


olasitarska:
Closing this, as I'm working on more improvements to Content Types.

PR #743, opened 2016-08-17
Stop _get_queued_tasks flushing task queues other than the one specified

RoganMurley:
Fixes #[742].

The internal testing function that gets a specific test stub task queue flushed all task queues instead of just the one specified. This caused a bug in `process_task_queues` where new tasks createdwhile processing a queue that do not belong to that queue would be flushed, and therefore never have a chance to execute.


adamalton:
Thanks Rogan!

PR #754, opened 2016-09-02
Deferred builtin check

mattdoughty:
The deferred builtin overrides Djangae's deferred handler, so deferred tasks will sometimes fail when it's switched on. This isn't obvious to spot, so I've added a system check which will warn you when the builtin is on in app.yaml.

Had to add PyYAML and mock to the testapp requirements so I could change app.yaml in the tests - hope this is ok.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### New features & improvements:
 
 - Cleaned up and refactored internal implementation of SimulatedContentTypeManager. Now also allows patching ContentType manager in migrations. 
+- System check for deferred builtin which should always be switched off.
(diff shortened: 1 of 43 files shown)


adamalton:
Hey MAtt,

* Do we need to add `yaml` as a requirement?  Can we just use the one that's bundled with the App Engine SDK, or does that not work?
* Can we swap `mock` for `djangae.contrib.sleuth`?  It just helps to keep the dependencies limited to `django` and the GAE SDK.

mattdoughty:
Didn't know about sleuth! Have done that and removed the pyyaml dependency.

PR #759, opened 2016-09-21
Fix dumpurls

adamalton:
Previously, when running the `dumpurls` management command, it would error with `AttributeError: 'module' object has no attribute 'views'` because `admindocs.views` had not been imported.

This fixes that.

PR checklist:
- [-] Updated relevant documentation
- [x] Updated CHANGELOG.md 


asendecka:
Looks good.

PR #760, opened 2016-09-21
Log invalid sender address

### LINKED ISSUE #758: Log email address when InvalidSenderError is raised
When App Engine raises `InvalidSenderError` (when you try to send an email with a disallowed `from` address) it does not log the email address which you are trying to send from.  This is incredibly annoying.

We should catch this exception, log the `from` address, and then re-raise it.  In `djangae/mail.py`, we currently have this:

```python
try:
        message.send()
    except (aeemail.Error, apiproxy_errors.Error):
        if not fail_silently:
            raise
```

which we should change to something like this:

```python
from google.appengine.api.mail_errors import InvalidSenderError

try:
        message.send()
    except (aeemail.Error, apiproxy_errors.Error) as e:
        if isinstance(e, InvalidSenderError):
            logging.warning("Invalid 'from' address: %s", message.from)  # or whatever it is
        if not fail_silently:
            raise
```

Note that this same code appears in 2 places in this file.

adamalton:
Fixes #758.

- Added basic tests for the Djangae email backends.
- Slightly refactored code in the email backend to avoid a bit of duplicated code.
- Added logging of the 'from' address when App Engine rejects it with `InvalidSenderError`.

PR checklist:
- [-] Updated relevant documentation
- [x] Updated CHANGELOG.md 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Cleaned up and refactored internal implementation of SimulatedContentTypeManager. Now also allows patching ContentType manager in migrations.
 - Add ability to specify GAE target instance for remote command with `--app_id` flag
+- When App Engine raises an `InvalidSenderError` when trying to send an email, Djangae now logs the 'from' address which is invalid (App Engine doesn't include it in the error).
 
 ### Bug fixes:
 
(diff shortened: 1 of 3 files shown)


asendecka:
Looks good to me.

adamalton:
🌈

PR #763, opened 2016-09-23
Use serving url for Cloud Storage images

### LINKED ISSUE #748: Cloud Storage backend: use `get_serving_url` instead of the Cloud Storage bucket URL in `url` property
...if the file is an image.

Currently you'd have to manually do this with the Cloud Storage key.

scottg1989:
Fixes #748 

Changed the cloud storage url method so that it gets the serving url. This now returns the correct path for images, and also makes use of the cache which stores the serving urls.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - Fixed `./manage.py runserver` not working with Django 1.10 and removed a RemovedInDjango110Warning message at startup.
 - Restore `--nothreading` functionality to runserver (this went away when we dropped support for the old dev_appserver)
 - Fixed a bug where the `dumpurls` command had stopped working due to subtle import changes.
+- Utlise `get_serving_url` to get the correct url for serving images from Cloud Storage
 
 ### Documentation:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 - Fixed `./manage.py runserver` not working with Django 1.10 and removed a RemovedInDjango110Warning message at startup.
 - Restore `--nothreading` functionality to runserver (this went away when we dropped support for the old dev_appserver)
 - Fixed a bug where the `dumpurls` command had stopped working due to subtle import changes.
-- Utlise `get_serving_url` to get the correct url for serving images from Cloud Storage
+- Utilise `get_serving_url` to get the correct url for serving images from Cloud Storage
 
 ### Documentation:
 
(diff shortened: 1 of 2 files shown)


adamalton:
This has a couple of side effects:

1. If you have a model with an `ImageField` on it, then `get_serving_url` is now called each time you save that object, meaning that if you modify that object in a transaction it must be a cross-group transaction.
2. It seems to cause errors in tests, because the local server is not running, and so the calls to `cloudstorage.stat` (which locally make a call to `http://localhost:8080/_ah/gcs`) get a 404.  My analysis of this might be wrong, but I think that's what's happening, traceback below.

```python
ERROR: controlpanel.tests.test_things:ThingEditingTests.test_increasing_monkey_increases_monkey_and_bananas_waiting_list
  vi +291 src/controlpanel/tests/test_things.py  # test_increasing_monkey_increases_monkey_and_bananas_waiting_list
    thumbnail = mommy.make(ThingImage, image_type="THUMBNAIL")
  vi +26  src/jungle/test.py  # wrapper
    return self._original(model, *args, **kwargs)
  vi +79  third_party/local/sitepackages/model_mommy/mommy.py  # make
    return mommy.make(**attrs)
  vi +262 third_party/local/sitepackages/model_mommy/mommy.py  # make
    return self._make(commit=True, **attrs)
  vi +321 third_party/local/sitepackages/model_mommy/mommy.py  # _make
    return self.instance(model_attrs, _commit=commit)
  vi +340 third_party/local/sitepackages/model_mommy/mommy.py  # instance
    instance.save()
  vi +195 src/things/models.py  # save
    self.image_url = self.image_file.url
  vi +69  src/sitepackages/django/db/models/fields/files.py  # _get_url
    return self.storage.url(self.name)
  vi +304 src/sitepackages/djangae/storage.py  # url
    url = get_serving_url(self._get_blobkey(filename))
  vi +308 src/sitepackages/djangae/storage.py  # _get_blobkey
    blob_key, info = _get_or_create_cached_serving_url(self._add_bucket(name))
  vi +94  src/sitepackages/djangae/storage.py  # _get_or_create_cached_serving_url
    info = cloudstorage.stat(blob_key_or_info)
  vi +151 src/sitepackages/cloudstorage/cloudstorage_api.py  # stat
    body=content)
  vi +132 src/sitepackages/cloudstorage/errors.py  # check_status
    raise NotFoundError(msg)
NotFoundError: Expect status [200] from Google Storage. But got status 404.
Path: '/app_default_bucket/mock-img.jpeg'.
Request headers: None.
Response headers: {}.
Body: ''.
Extra info: None.
```

adamalton:
Update: the first one is now fixed in #765.  The second one is caused when you try to access `.url` on a file which hasn't been saved yet.  It's just that the error message doesn't make that clear.

Maybe we should catch the `cloudstorage.NotFoundError` and raise a better message.

PR #765, opened 2016-09-26
Make calls to get serving url non transactional

adamalton:
Fixes one of the side effects of #763, which is that calling `get_serving_url` causes a Datastore read, and therefore if you access `my_model_instance.image_field.url` inside a transaction it adds an entity group to that transaction.

- In both the `CloudStorage` and `BlobstoreStorage`, wrap calls to `get_serving_url` in `transaction.non_atomic`.

PR checklist:
- [X] Updated CHANGELOG.md 
- [X] Added tests for my change


asendecka:
Looks good to me!

adamalton:
Thanks Ola!

PR #777, opened 2016-11-14
Handle LargeImageError in BlobstoreStorage.url()

asendecka:
url() method was not handling big files (>32MB) and broke when the `get_serving_url` was throwing `LargeImageError`

- catch `LargeImageError` in try clause when we attempt `get_serving_url` 

PR checklist:
- [x] Updated CHANGELOG.md 


adamalton:
I've created #779 as a follow on from this.  I'm going to merge this as it's a definite improvement on the current situation.

PR #782, opened 2016-11-17
Update user email address in gauth middleware if it gets changed

### LINKED ISSUE #781: Auth backend doesn't update user emails if they change their primary Google account email
1. User logs into app with their google account which has a primary email of bob@cat.com
2. Auth backend creates a `User` object
3. User changes their primary address on their google account to bob@gmail.com and logs into app again
4. User logs in successfully but `User.email` is still bob@cat.com

It looks like this behaviour may have changed in #722 where the logic to update emails in the auth middleware was removed in favour of it happening in the backend. However this doesn't seem to be happening.

An existing user will never get past the below short circuit.
```
        try:
            # User exists and we can bail immediately. With one caveat.
            user = User.objects.get(username=user_id)
            # Ensure that the user has got both the `email` and `email_lower` fields populated
            if not user.email_lower:
                # The user existed before the introduction of the `email_lower` field. Update it.
                user.email = email
                # The save will also update the email_lower field
                user.save(update_fields=['email', 'email_lower'])
            return user
        except User.DoesNotExist:
            pass
```

nealedj:
Fixes #781 .

- Fixed auth middleware to update user email address if it gets changed

PR checklist:
- [x] Updated CHANGELOG.md 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - Fixed a side effect of that ^ introduction of `get_serving_url` which would add an entity group to any transaction in which it was called (due to the Datastore read done by `get_serving_url`).
 - Fixed fetching url for non images after introduction of `get_serving_url` call inside CloudStorage url method.
 - Fixed fetching url for files after introduction of `get_serving_url` call inside BlobstoreStorage url method when file is bigger than 32MB.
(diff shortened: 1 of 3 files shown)


adamalton:
Woo!  🌈

Issue #791, opened 2016-12-05
Simulated ContentTypes & Django admin LogEntry


philoozushi:
Using current Djangae scaffold, and Djangae v0.9.7+ :
In Django admin "Recent Actions" box, all logs display "Unknown content" for the content-type.

I don't get how exactly the simulated ContentTypes work, but in Django shell, I can see a correct `content_type_id` for LogEntries, matching an existing `ContentType`, however the ForeignKey `content_type` field raises an error.

```
>>> entry = LogEntry.objects.last()
>>> entry.content_type_id
-5352591944405510358L
>>> ContentType.objects.get(id=-5352591944405510358L)
<ContentType: MyModel>
>>> entry.content_type
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "...sitepackages/django/db/models/fields/related_descriptors.py", line 169, in __get__
    rel_obj = qs.get()
  File "...sitepackages/django/db/models/query.py", line 387, in get
    self.model._meta.object_name
DoesNotExist: ContentType matching query does not exist.
```

adamalton:
This is the bit that looks suspicious:

```pythong
>>> entry.content_type_id
-5352591944405510358L
```

I think this is probably the same underlying issue as this: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/djangae-users/HHiJCqGxeAc/aOP6qDXdAwAJ

It seems that we're getting negative content type IDs, which is clearly wrong.  The obvious culprit is this line: https://github.com/potatolondon/djangae/blob/427a4f34a9a58b9a7dabd899fe1c5564c0969f6b/djangae/contrib/contenttypes/models.py#L38

adamalton:
If you could write a test that produces the issue then that would be incredibly useful in helping to get it fixed.

philoozushi:
Sure, I though it was suspicious too.
I added a test to check if ContentType ids are positive.

PR #792, opened 2016-12-05
SimulatedContentTypeManager._get_id test cf. #791

philoozushi:
Adding test for #791.

### DIFF
--- a/djangae/contrib/contenttypes/tests.py
+++ b/djangae/contrib/contenttypes/tests.py
@@ -41,6 +41,13 @@ class SimulatedContentTypesTests(TestCase):
         self.assertEqual(by_id.model, DummyModel._meta.model_name)
         self.assertEqual(by_id.app_label, DummyModel._meta.app_label)
 
+    def test_ids_are_positive(self):
+        # actually testing SimulatedContentTypeManager._get_id provide correct ids
+        # cf. https://github.com/potatolondon/djangae/issues/791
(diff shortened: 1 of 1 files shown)


adamalton:
Thanks.  I'm going to leave @Kazade to look at this, or to at least explain the CRC stuff to me and why it's shifted down by `2**63`.

philoozushi:
For information, with my project, this repo test app, or Djangae scaffold app, I always have these kind of ids :
```python
>>> for ct in ContentType.objects.all(): print ct.id, ct.app_label, ct.model
... 
-6311814338746989887 contenttypes contenttype
8785559794874550405 djangae group
7361026521869113233 cspreports cspreport
-3767371066152500551 auth permission
5765535199569929304 auth group
4645563115640020792 djangae countershard
-3961434392430070340 admin logentry
4840040022415803340 sessions session
7403309253764677893 djangae gaedatastoreuser
```
If it helps...

adamalton:
Thanks.

So we had a conversation about this, and the negative IDs are actually deliberate.  The Datastore only allows you to store signed integers, and so we have to use the full range (from minus 2**63 to plus 2**63) in order to accommodate all possible values.

The problem then comes when you try to save one of these to a field that expects them to be positive.

How we "fix" it without making a backwards incompatible change that affects existing projects we're not quite sure.  Thinking in progress...

PR #798, opened 2016-12-15
Add weak (memcache) locking

nealedj:
- Add soft (memcache) locking to contrib.locking

PR checklist:
- [] Updated relevant documentation
- [x] Updated CHANGELOG.md 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### New features & improvements:
 
--
+- Implemented weak (memcache) locking to contrib.locking
 
 ### Bug fixes:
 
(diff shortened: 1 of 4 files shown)


adamalton:
Ooh!  Nice.

Looks like the `acquire` method on the `MemcacheLock` class shouldn't have have the `kind` kwarg, right?

nealedj:
Ah yeah, totally pointless parameter. Consider it removed.

nealedj:
FYI this is based on the ghetto lock described in the [memcached docs](https://github.com/memcached/memcached/wiki/ProgrammingTricks#ghetto-lock).

adamalton:
Cool.  I think that the abuse of the memcache expiry time as the `steal_after_ms` is acceptable.  There's an edge case that if you have 2 places which both use the same key but have different `steal_after_ms` times, then it won't work as expected.  But I'm not sure if it's worth having a significantly more complex implementation in order to remove that edge case.

Might be worth adding a test for the `steal_after_ms` stuff though, just to test that it at least works as far as this implementation is meant to.

And maybe we should add a quick note to the docs about that ^ edge case?

PR #804, opened 2017-01-16
Fix content type relations

### LINKED ISSUE #791: Simulated ContentTypes & Django admin LogEntry

Kazade:
Fixes #791 
Closes #792 

- Makes use of the SimulatedContentTypes manager in reverse lookups via ForeignKey
- Replaces the ContentType.pk field with a BigIntegerField to prevent validation issues

PR checklist:
- [x] Updated CHANGELOG.md 


adamalton:
@Kazade I realised that we haven't add the migration for this.  So I looked into it... In djangae/apps.py, we have this:

```python
conn = connections[router.db_for_read(ContentType)]
if conn.settings_dict.get("ENGINE") != 'djangae.db.backends.appengine':
    # Don't enforce djangae.contrib.contenttypes if content types are being
    # saved to a different database backend
    return
```

So if you're using SQL for ContentTypes then we don't force you to use **djangae** contettypes.  But we also don't prevent it.  So I'm thinking that we probably should add the migration, because if anyone has that setup then it's definitely not going to work currently.

Thoughts?

PR #814, opened 2017-01-25
770 auth reorganize

grzes:
Fixes #770 

We decided that having an app named `sql` isn't the best idea so we renamed the gauth authentication apps to gauth_datastore and gauth_sql. This is backwards incompatible and will require upgrading apps to replace `gauth.datastore` to `gauth_datastore` and any mention of `contrib.gauth.common` to just `contrib.gauth`. (gauth.sql shouldn't be an issue cause it never quite worked).

All datastore models have meta `db_table` altered to ease the transition.


farridav:
Nice! Id be happy to see this in, then we can close https://github.com/potatolondon/djangae/pull/770

PR #815, opened 2017-01-25
Introduce support for migrations on the datastore

### LINKED ISSUE #438: Implement Migrations
With the new (as of 1.7) Django migrations stuff, we could potentially implement some stuff to allow "migrations" (in a slightly different way) to be run.

The basic idea would be to allow mapreduce-type tasks to be run to alter data in the DB.  We could largely do this with `djangae.contrib.mappers.defer_iteration`, but we would need to know when the mappers had finished, which we can't currently do with that.

We could also/instead potentially implement something similar to Django's `RunPython` thing, which would run stuff on the Datastore, but (due to the fact that we can't do long-running tasks) it would be more like `RunThisOverEachEntity`.

We may be able to do some of the stuff by implementing the methods in our backend's `DatabaseSchemaEditor`.

Kazade:
Fixes #438 

- Add djangae.db.migrations
- Add a custom sharded datastore mapper for running migrations on low-level entities
- Implement several useful migration operations 

PR checklist:
- [x] Updated CHANGELOG.md 


Kazade:
OK here are some things left to do for this:

 - Implement sharding on keys which have names rather than IDs
 - Document the migration operations
 - Add start/finish timestamps to the tasks just for more information

@adamalton can you review what I've done so far, and then I'll have another go at finishing this off!

adamalton:
Nice, I'll take a look.  A few more things for the TODO list:

* See if we can fix the `_set_identifider` methods so that they can get hold of the migration name.  Otherwise our identifiers for the migration tasks are not guaranteed to be unique.  Can we use `__file__` or something?
* Should we add a migration operation that adds unique markers when populating a new field?  Or should we modify the `AddFieldData` one to do that?  People _could_ just use the uniquetool, but that means you can't manage it as part of the migration flow.
* The documention should mention that we don't currently support migrations that iterate over Django model instances, but that it's a possible future addition.

adamalton:
Bagsy reviewing this.

adamalton:
Note to self: we need to add a note in the documentation to say that if you use the `MapFunctionOnEntities` migration then you are responsible for doing your own transaction handling in the function that you pass.

olasitarska:
This is very exciting! ✨

Kazade:
OK, I think this is ready to merge. But I'll leave the honour of merging up to @adamalton

PR #826, opened 2017-02-03
Iterator field max length

### LINKED ISSUE #729: ListField & SetField max_items constraint
So it turns out that - contrary to beliefs previously held by the Djangae developers - the number of items in a ListField or SetField is not limited to 500.

The limit appears to be 20,000 minus the number of other indexed properties on the instance.  So I assume that each other "normal" field counts as one indexed property, each value in each other list/set field counts as one indexed property, and each compound index for the model in index.yaml counts as one indexed property.  Although that might not be right - what if a compound index contains a list field?

In any case, calculating how many items can be in any given list/set field and then somehow validating that in `full_clean` or `clean_fields` is going to be hard and would probably require a mixin.

What we should do is:
1. Find out how the limit is calculated (i.e. figure out whether or not compound indexes are included), and add this info to the documentation for `ListField` and `SetField`.
2. Add an optional `max_items` kwarg to `ListField` and `SetField` so that people can specify their own safe limits.  Make the fields validate based on that limit.
3. Possibly provide a validation mixin which automatically calculates/detects the maximum possible number of items.  Although this might not be worth doing, it depends on how complex the calculation is.

RoganMurley:
Fixes #729.

- Added max_length validation support to `ListField`, `SetField`, `RelatedListField` and `RelatedSetField`.

PR checklist:
- [*] Updated relevant documentation
- [*] Updated CHANGELOG.md 
- [*] Added tests for my change


adamalton:
@RoganMurley (or anyone!) could you please have a look at my changes, and merge if you think they look good.

RoganMurley:
These look good to me!

adamalton:
\0/

PR #830, opened 2017-02-09
Fix nested RelatedSetField bug.

### LINKED ISSUE #829: KeyError on nested prefetching of RelatedSetField
Continuing from the fix to #774, there's a bug with nested keys.

```python
# =======  Models  ======= 
class Post(models.Model):
    content = models.TextField()
    authors = fields.RelatedSetField('Author', related_name='posts')

class Author(models.Model):
    name = fields.CharField()
    awards = fields.RelatedSetField('Awards')
    
class Awards(models.Model):
    name = fields.CharField()
    
# =======  Usage  ======= 
# Causes a KeyError
posts = Post.objects.all().prefetch_related('authors__awards')
```

This happens because the values being read from the nested field come back as long, which does not match what's in `matchers` in #808. Fix incoming in a PR.

mvd7793:
Fixes #829.

### DIFF
--- a/djangae/fields/related.py
+++ b/djangae/fields/related.py
@@ -146,7 +146,9 @@ class RelatedIteratorManagerBase(object):
                 was passed in so we can convince Django to do the right thing when merging
             """
             for related_thing in queryset:
-                for i, matcher in enumerate(matchers[related_thing.pk]):
+                # Allow nested prefetching from other RelatedSetFields
+                pk = int(related_thing.pk)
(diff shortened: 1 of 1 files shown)


moggers87:
A quick test would be nice ☺

Kazade:
I also mentioned this in #829 but basically we had a bug before which was causing IDs to not be saved as integers which would cause this bug (the casting to integer in this PR just works around it). Hopefully this will go away now with a resave on latest Djangae.

PR #833, opened 2017-02-10
Grzes 770 auth reorganize

adamalton:
This is just #814 with a couple of changes to the CHANGELOG.


adamalton:
Closes #814, #770

PR #840, opened 2017-02-15
Clean up the result fetching code

Kazade:
This results in a number of improvements/changes:

 - Queries are now fetched in cursor.execute() not during iteration which is more like other backends. This has the nice side-effect that Django Debug Toolbar now correctly shows query times
 - Results are not wrapped in generators, making tracebacks from Djangae/AppEngine much more readable
 - SelectCommand.execute() now returns the number of returned results, so cursor.rowcount is correctly set
 - Transforms which are applied to entities are now outside _fetch_results() making the code lots more readable, and also makes profiling easier
 - Performance should be marginally improved due to fewer variables / function calls

PR checklist:
- [x] Updated CHANGELOG.md 


asendecka:
Looks good to me. Happy to see `while` loop gone and the readability of the code improved immensely. 👍

PR #846, opened 2017-02-22
Updated AppEngineSecurityMiddleware to work with Django >= 1.10

jhovanic:
Django middleware changed in 1.10. See more [here](https://docs.djangoproject.com/en/1.10/topics/http/middleware/#upgrading-pre-django-1-10-style-middleware)


Kazade:
Thanks!

PR #847, opened 2017-02-23
Detect when user is switching to the same account

mattdoughty:
Currently if a user clicks on switch account but then decides to use the same account, they'll get logged out. If they then try to login again, they'll get redirected to `switch_accounts` which will log them out again. The user will then get stuck in a loop with no way of getting back into your app from the same tab - they'll have to open a new tab, sign into their Google account and then open your app.

The way I've partially solved this is to add a session variable for storing if the user already been redirected to `create_logout_url`. It'll work like this:

User is signed in to a single Google account
- User goes to `switch_accounts` view.
- `switch_accounts` redirects to `create_login_url`.
- User is signed in to one Google account, so returns to `switch_accounts`.
- `switch_account` detects they've returned with the same account so redirects to `create_logout_url`. It'll set a session variable that it's done so.
- User enters the same username and password in, goes back to `switch_accounts`
- `switch_accounts` detects they've returned with the same account, but also checks the session variable. As the session variable is already set, we're done and redirect to final destination.

User is signed in to multiple Google accounts. This is more difficult to do properly because when the user returns with the same account, we don't know if that's because they've chosen the same account or they're only signed in with one account. We could do this by measuring how long it took to get redirected, but this would probably be unreliable. What I've done here is an improvement, however, as the user selecting the same account would get logged out, but could then enter the same username and password and return to the app safely.

### DIFF
--- a/djangae/contrib/gauth/tests.py
+++ b/djangae/contrib/gauth/tests.py
@@ -477,6 +477,39 @@ class SwitchAccountsTests(TestCase):
             self.assertEqual(len(expected_user_query), 1)
             self.assertEqual(int(self.client._session()['_auth_user_id']), expected_user_query[0].pk)
 
+    @override_settings(DJANGAE_CREATE_UNKNOWN_USER=True)
+    def test_switch_to_same_account(self):
+        gcu = 'djangae.contrib.gauth.middleware.users.get_current_user'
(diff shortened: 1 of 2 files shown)


Kazade:
Thanks for fixing this!

PR #854, opened 2017-03-08
Allow additional validators to be specified on djangae.fields.CharField

### LINKED ISSUE #850: fields.CharField wipes out inline validators
When using [fields.CharField](https://github.com/potatolondon/djangae/blob/f2a2240c5f6b3b67baf02c6ce7e284524f619d2f/djangae/fields/__init__.py#L86
) any inline validators (e.g `field.CharField(validators=[validate_my_thing])`) are wiped out because of the assignment here https://github.com/potatolondon/djangae/blob/f2a2240c5f6b3b67baf02c6ce7e284524f619d2f/djangae/fields/__init__.py#L86

Should be an easy fix (appending)

Kazade:
Fixes #850 

- Fixes a bug where the Djangae CharField would get its validators overridden

PR checklist:
- [x] Updated CHANGELOG.md 


farridav:
quick turnaround! thanks @Kazade @scottg1989

PR #869, opened 2017-03-27
Improve contains query indexing

### LINKED ISSUE #730: Change indexing behaviour for `__contains` on CharField
Currently, when you do an `__contains` filter on a `CharField`, Djangae's special indexing adds one or several index columns to store all the possible substrings.

There are 2 things wrong with this:
1. If there are more than 500 different substrings then it uses multiple columns to store them.  This is based on the incorrect idea that list fields can contain a maximum of 500 items.  It turns out that that's not true: #729.
2. The `number_of_permutations` method (which calculates the number of substrings) doesn't take account of duplicates.  E.g. for the string `"aba"` it assumes that there are 6 substrings (`["a", "ab", "aba", "ba", "b", "a"]`), but actually there are only 5 distinct substrings.

The maximum amount of text that we can index is limited by:
1. The `google.appengine.api.datastore._MAX_INDEXED_PROPERTIES` (see #729).
2. The fact that an entity can be a maximum of 1MB.
3. The fact that the maximum indexable single string value is 1500 bytes (`google.appengine.api.datastore_types._MAX_STRING_LENGTH`).

We should be able to scrap both the `MAX_COLUMNS_PER_SPECIAL_INDEX` setting and our `CHARACTERS_PER_COLUMN` variable.

Kazade:
Fixes #730

This introduces a new method of storing indexed data for special indexes which solves a number of problems.

Currently, `contains` indexing generates a large amount of index data in a list field in the owning entity. This makes the entities "heavy" and slows down accesses across the site as more data has to transfer over the RPC, on top of that, we hit various restrictions as the number of permutations grows, and although we *could* index less data, the subsequent query would result in an inequality.

This method creates an entity for each contains/icontains index on a field. These index entities are descendants of the owning entity and so don't use up entity groups in transactions. As the owning entity key is part of the index entity key, we can quickly gather the entities that match the query by querying for matching index entities. This adds an additional query to every __contains query and an additional put to every entity which has a 'contains' index but I believe the benefits of this method outweigh those costs.

Left to do:

 - Write documentation mentioning restrictions

- Move indexed data for 'contains' and 'icontains' queries to their own entities which are descendants of the owning entity

PR checklist:
- [x] Updated CHANGELOG.md 


adamalton:
This looks good to me, I've just got a few things that I want to check with @Kazade before I merge it.

adamalton:
MERGING!! 🌻

PR #874, opened 2017-03-31
Processing module

### LINKED ISSUE #407: Mappers don't work on models with non-integer PKs
I discovered this when using `djangae.contrib.uniquetool` (which uses `djangae.contrib.mappers`), and I think it's an issue with the mappers stuff.

If your model's primary key is not an integer-ish field then you get this error:

```
Internal Server Error: /_ah/mapreduce/kickoffjob_callback/1574160994066DC1D0E2C (/base/data/home/apps/s~my-project/57-df974ea24d-stable.386079706619870743/sitepackages/django/core/handlers/base.py:231)
  File "/base/data/home/apps/s~my-project/57-df974ea24d-stable.386079706619870743/sitepackages/django/core/handlers/base.py", line 111, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/base/data/home/apps/s~my-project/57-df974ea24d-stable.386079706619870743/sitepackages/django/views/decorators/csrf.py", line 57, in wrapped_view
    return view_func(*args, **kwargs)
  File "/base/data/home/apps/s~my-project/57-df974ea24d-stable.386079706619870743/sitepackages/djangae/utils.py", line 173, in request_handler_wrapper
    view_func.dispatch()
  File "/base/data/home/runtimes/python27/python27_lib/versions/third_party/webapp2-2.3/webapp2.py", line 547, in dispatch
    return self.handle_exception(e, self.app.debug)
  File "/base/data/home/runtimes/python27/python27_lib/versions/third_party/webapp2-2.3/webapp2.py", line 545, in dispatch
    return method(*args, **kwargs)
  File "/base/data/home/apps/s~my-project/57-df974ea24d-stable.386079706619870743/sitepackages/mapreduce/base_handler.py", line 131, in post
    self.handle()
  File "/base/data/home/apps/s~my-project/57-df974ea24d-stable.386079706619870743/sitepackages/mapreduce/handlers.py", line 1365, in handle
    readers, serialized_readers_entity = self._get_input_readers(state)
  File "/base/data/home/apps/s~my-project/57-df974ea24d-stable.386079706619870743/sitepackages/mapreduce/handlers.py", line 1437, in _get_input_readers
    readers = input_reader_class.split_input(split_param)
  File "/base/data/home/apps/s~my-project/57-df974ea24d-stable.386079706619870743/sitepackages/djangae/contrib/mappers/readers.py", line 76, in split_input
    pk_range = last_id - first_id
TypeError: unsupported operand type(s) for -: 'unicode' and 'unicode'
```

### LINKED ISSUE #506: mapreduce: sharding issues
There’s a couple issues with your split_input implementation:
(1) I’m not finding balanced distribution across shards. Some shards are even totally unused. It's possible this may occur for older applications that include older model entities that used the old sequential id strategy rather than the newer scattered one. But I'm not sure if there could be other reasons as well.
(2) It requires a -pk datastore index for the model, which official mapreduce AbstractDatastoreInputReader does not.

It looks like the official mapreduce library implements its shard splitter by ordering by a model property called **scatter** and sampling datastore keys, which it then uses to generate a key range for each shard. Using a Google Search, I discovered that this **scatter** property is used to get random distributed samples in the keyspace. The lowest key range has no starting key, and the highest key range has no ending key. Perhaps we should do the same: https://github.com/GoogleCloudPlatform/appengine-mapreduce/blob/71ed723e494dbcbe3ea77220612b938abeb452a5/java/src/main/java/com/google/appengine/tools/mapreduce/inputs/DatastoreShardStrategy.java#L58

Maybe a solution would be to use context manager when querying for samples to notify the SelectCommand to order by **scatter**, as long as there are no inequality filters.

### LINKED ISSUE #509: mapreduce: need finalized callback
DjangaeMapperPipeline.callback is not the final callback to be made. DjangaeMapperPipeline.finalized is, as it contains the output counters, whereas the output counters have not yet been filled when calling DjangaeMapperPipeline.callback.

Also, it'd be great for either callback to the task to include the job name.

### LINKED ISSUE #510: Discussion - How we should handle MapReduce and Pipelines
Apologies in advance for this massive wall of text

OK so i've been looking at how we should handle data processing in djangae on and off for a while now and every time I sit down and look at it I come to different conclusions.
# The situation

MapReduce is essentially a wrapper around the Pipelines library, which allows generic data processing asynchronous functionality in a similar manner to the tasks queues, only with POWER. Here are an awesome examples stolen from the Pipeline Wiki, you can see the full code at 

https://github.com/GoogleCloudPlatform/appengine-pipelines/blob/master/python/demo/main.py

and read more about Pipelines at

http://sookocheff.com/series/pipelines-api/

```
class LongCount(pipeline.Pipeline):

  def run(self, entity_kind, property_name, value):
    cursor = None
    count = 0
    while True:
      query = db.GqlQuery(
          'SELECT * FROM %s WHERE %s = :1' % (entity_kind, property_name),
          value.lower(), key_only=True, cursor=cursor)
      result = query.fetch(1000)
      count += len(result)
      if len(result) < 1000:
        return (entity_kind, property_name, value, count)
      else:
        cursor = query.cursor()

class SplitCount(pipeline.Pipeline):

  def run(self, entity_kind, property_name, *value_list):
    all_counts = []
    for value in value_list:
      stage = yield LongCount(entity_kind, property_name, value)
      all_counts.append(stage)

    yield common.Append(*all_counts)

class CountReport(pipeline.Pipeline):

  def run(self, email_address, entity_kind, property_name, *value_list):
    yield common.Log.info('UselessPipeline.coolness = %s',
                          (yield UselessPipeline()).coolness)

    split_counts = yield SplitCount(entity_kind, property_name, *value_list)
    yield common.Log.info('SplitCount result = %s', split_counts)

    with pipeline.After(split_counts):
      with pipeline.InOrder():
        yield common.Delay(seconds=1)
        yield common.Log.info('Done waiting')
        yield EmailCountReport(email_address, split_counts)

  def finalized(self):
    if not self.was_aborted:
      logging.info('All done! Found %s results', self.outputs.default.value)
```

As you can see from the control flow in CountReport, with methods like After() and InOrder() you can really do some awesome things!

However the appengine pipelines library is a blessing and a curse. To get this to properly work with djangae we need to get the Pipeline requests to execute inside the django context, which in itself should not be that big a deal, we just have to make sure we route all the /_ah/pipelines traffic, we can see this in the Pipeline class

```
  def start(self,
            idempotence_key='',
            queue_name='default',
            base_path='/_ah/pipeline',
            return_task=False,
            countdown=None,
            eta=None):
```

If we do this then we _should_ have all the pipeline requests being executed in the django context, which is nice!

This problem is when we get to the MapReduce library....

Even though the Pipelines library is used they have rolled out their own handlers so there is a status dashboard which is specific to MapReduce, and the lovely pattern we can see with Pipelines is slowly destroyed. While I can see the idea behind this it also makes everything very hard to implement without re-writing a large portion of the library. 
# Suggested reading

Start with 
https://github.com/GoogleCloudPlatform/appengine-mapreduce/blob/master/python/src/mapreduce/mapreduce_pipeline.py 

Take a look at the MapPipeline class, this looks nice and inline with the pattern we expect from pipelines, you'll see there is a MapperPipeline which is yielded from this class. The pattern starts to break down at this point with a call to control.start_map which calls the Mapper Handler directly, I recently got a PR merged that allows us to override the base_path at this point, previously this was not possible so almost total code duplication was needed to get anywhere.
# In an ideal world

So what do we even want at this point? Well for MapReduce to work nicely with djangae we need it to support an InputReader that can handle Django models, this has already mostly been written. If we can coerce the entire MapReduce library to pass through the Django context as well as the Pipelines library then in theory we can just use the vanilla API, I think this is the best option and most of my adventures in this area have been trying to make this happen.

There is a lot to cover here with this problem and I myself can't even remember all the issues until I sit down and look at my attempts, if you would like to help then please have a look at the linked code samples and add some questions/suggestions to this issue if you have any and I will continue to add issues to this ticket as a thought dump.

TL;DR - Pipelines = Task Queues on steroids, MapReduce library = Pipelines in jail

### LINKED ISSUE #789: MapReduce tasks map through all namespaces when using the default DB
See https://github.com/GoogleCloudPlatform/appengine-mapreduce/blob/master/python/src/mapreduce/input_readers.py#L340

If the `NAMESPACE` passed in the mapper spec is `None` then the input reader will map over entities in all namespaces. When djangae sets up an MR pipeline it [takes the namespace](https://github.com/potatolondon/djangae/blob/master/djangae/contrib/mappers/pipes.py#L143) from the DB settings of the specified database. The initial [database settings](https://github.com/potatolondon/djangae/blob/master/djangae/settings_base.py#L8) in settings_base don't set the `NAMESPACE` and therefore we always pass `None` through to the pipeline and map over all entities.

If this is an issue I guess we could fix this in a couple of ways:

- Add `NAMESPACE: ''` to settings_base
- Change fetches of `NAMESPACE` to default to empty string. This would need doing in `MapReduceTask` and `RawMapperMixin` in the uniquetool (possibly elsewhere too...?)

Kazade:
(I think all of the below have been fixed, reopen if that's not the case)

Fixes #789 
Fixes #506
Fixes #509
Fixes #510 
Fixes #407

- Adds `djangae.contrib.processing.mapreduce`
- Replaces the URLs from `djangae.contrib.mappers`
- Provides the start of a new API to mapreduce
- Reimplements `djangae.contrib.mappers` in terms of the new app

PR checklist:
- [x] Updated CHANGELOG.md 


grzes:
This looks good, BUT!

the tests weren't ran at all and after adjusting our test runner (in master) they die with the `Cannot read more than the available bytes....` error, which is covering up an `UnexpectedPipelineError Received evaluation task for pipeline ID on attempt 2, which will not be ready until...` error.

PR #888, opened 2017-04-09
change typo on CharField

oonid:
to store large char values, use text field, not this (CharField) field.

- only change one word to avoid confusing on documentation.

PR checklist:

### DIFF
--- a/docs/fields.md
+++ b/docs/fields.md
@@ -176,7 +176,7 @@ This field requires no special kwargs, and should accept all standard Django fie
 
 ## CharField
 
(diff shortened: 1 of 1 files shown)


adamalton:
Thanks for raising this.

I think that the original version is actually deliberate.  The suggestion is to use `djangae.fields.CharField` instead of the usual `django.db.models.CharField`, because the Djangae version validates the number of _bytes_ in the value, rather than the number of characters (which are not necessarily the same).  So if you're storing strings which might be getting close to the limit, then using Djangae's `CharField` ensures that the validation correctly checks the number of _bytes_, thereby avoiding having your strings trimmed by the Datastore.

It could probably be worded better in the documentation, but I'm 99% sure that the current wording is not a typo.

Adam

PR #893, opened 2017-04-13
Improve caching performance

### LINKED ISSUE #831: Memcache layer slows down large serialization significantly
For sample models such as the following, with ~100 of each model, let's say we run the query shown, knowing we intend to serialize all data.

```python
# =======  Models  ======= 
class Post(models.Model):
    content = models.TextField()
    authors = fields.RelatedSetField('Author', related_name='posts')

class Author(models.Model):
    name = fields.CharField()
    awards = fields.RelatedSetField('Awards')
    
class Awards(models.Model):
    name = fields.CharField()

# =======  Query  ======= 
posts = Post.objects.all().prefetch_related('authors').prefetch_related('authors__awards')
# ...
# Perform a cheap operation that causes the Posts iterator to run.
# ...
serialize(posts)
```

Note that to run this code, you need the fix in #830. Given some significant number of models, we would expect most time to occur in fetching if the rest of the data on the models is simply in-memory:

```
Posts fetch (expensive)
|
|---> Author fetch (expensive)
      |
      |---> Tags fetch (expensive)
       		|
       		|---> Serialization (cheap)
```

What I've found in testing is that serialization is rather expensive because of memcache calls. I happen to be doing nested serialization with [Django Rest Framework](http://www.django-rest-framework.org/), which will traverse these links for you, but since nested objects are prefetched it should be cheap. Instead, I've found serialization takes 4000ms. A deep dive showed hundreds of `memcache.set` RPCs, which I traced to `KeyPrefixedClient.set_multi_async`. There are a couple problems here. 

1. `KeyPrefixedClient.set_multi_async` doesn't respect the `DisableCache` decorator, so it's impossible to disable these synchronous `set` calls.
2. The `set` calls are all done synchronously even though they're not required, and there isn't a way to batch them.

Fixing the first problem brings serialization down to 2000ms, which still isn't great, but debugging that is a separate issue.

### LINKED ISSUE #886: Context cache appears slow for many entities under load
In some cases I'm seeing up to 5 seconds to add 20+ large objects into the context cache. This becomes most apparant under load.

In this example from cloud trace the gaps A->B and C->D show the issue where: A/C is the end of the datastore query and B/D is the final put to memcache after the context cache has been filled.

trace

In our case the entities involved have many (20+) properties but are not particularly large. It's also possibly this issue is specific to our app, but can't see how ATM. We fixed this temporarily by disabling cache puts for the slow cases and reducing `max_concurrent_requests` to 2 per GAE instance.

A short-term solution would be to fix #885

Kazade:
Fixes #886 
Fixes #831 (hopefully)

- Heavily optimise the context cache which was copying wastefully
- Add a memory limit to the context cache (configurable by `DJANGAE_CACHE_MAX_CONTEXT_SIZE`)
- Restrict the number of entities which are cached after a PK query (configurable with `DJANGAE_CACHE_MAX_ENTITY_COUNT`)
- Fix a typo in the `DJANGAE_MAX_QUERY_BRANCHES` setting lookup

This PR needs some tests for the `CacheDict` class (particularly around entity eviction), but I'm going on holiday so feel free to pick this up @adamalton  :)

PR checklist:
- [x] Updated CHANGELOG.md 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
  - Switched the default storage backend (in settings_base.py) to cloud storage. If you need to retain compatibility make sure you
  override the `DEFAULT_FILE_STORAGE` setting to point to `'djangae.storage.BlobstoreStorage'`.
  - Added AsyncMultiQuery as a replacement for Google's MultiQuery (which doesn't exist on Cloud Datastore).  This is the first step towards support for Cloud Datastore and therefore Flexible Environment.
+ - Added a configurable memory limit to the context cache, limited the number of instances cached from query results and corrected `disable_cache` behaviour.
 
 ### Bug fixes:
 
--- a/djangae/db/backends/appengine/caching.py
+++ b/djangae/db/backends/appengine/caching.py
@@ -218,6 +218,12 @@ def add_entities_to_cache(model, entities, situation, namespace, skip_memcache=F
     if not CACHE_ENABLED:
         return None
 
+    context = get_context()
+
+    if not (context.context_enabled or context.memcache_enabled):
+        # Don't cache anything if caching is disabled
+        return
+
     # Don't cache on Get if we are inside a transaction, even in the context
     # This is because transactions don't see the current state of the datastore
     # We can still cache in the context on Put() but not in memcache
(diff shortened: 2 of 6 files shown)


adamalton:
🌻  sweet!

PR #901, opened 2017-04-25
Fix issue where serialization of IterableFields resulted in invalid JSON

### LINKED ISSUE #881: Deserialized SetField cannot be loaded back with loadData
If you have a model with a `SetField`, export the model with `dumpdata` in JSON format, and then try to insert back into the datastore with `loaddata`, a `DeserializationError` is raised which halts the process.

```
django.core.serializers.base.DeserializationError: Problem installing fixture 'data.json': Unable to parse string into iterable field: field_value was '[u'es']'
```

From quickly looking at the code it looks like we deserialized the `SetField` into a a string containing a list...

```
def value_to_string(self, obj):
    """
    Custom method for serialization, as JSON doesn't support serializing sets.
    """
    return str(list(self._get_val_from_obj(obj)))
```

But when we load the data, the `to_python` special casing to transform the string into a set or list fails, as the field type is a set but the string does not start with `{`, so no transformation occurs.

```
if (self._iterable_type == set and value.startswith("{") and value.endswith("}")) or \
    (self._iterable_type == list and value.startswith("[") and value.endswith("]")):
        value = [x.strip() for x in value[1:-1].split(",") ]
```

nealedj:
Fixes #881.

This may get superseded if #555 gets completed but it seemed worth fixing the short term problem.

- Fix issue where serialization of IterableFields resulted in invalid JSON

PR checklist:
- [x] Updated CHANGELOG.md 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
  - Accessing the Datastore from outside tests will no longer throw an error when using the test sandbox
  - The in-context cache is now reliably wiped when the testbed is initialized for each test.
  - Fixed an ImportError when the SDK is not on sys.path.
-
+ - Fix issue where serialization of IterableFields resulted in invalid JSON
 
 ## v0.9.9 (release date: 27th March 2017)
 
--- a/djangae/fields/iterable.py
+++ b/djangae/fields/iterable.py
@@ -1,5 +1,6 @@
 # STANDARD LIB
 import copy
+from decimal import Decimal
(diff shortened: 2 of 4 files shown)


grzes:
Added 3 more tests (https://github.com/potatolondon/djangae/commit/1c3d7adb46359590ec26b4559c8c8e10ce2537ad) if we can make them pass let's definitely grab this over the 902 solution.

PR #902, opened 2017-04-26
Json set field serialization fix

### LINKED ISSUE #881: Deserialized SetField cannot be loaded back with loadData

grzes:
Builds upon David's 555 branch - fixing up some code and changing the tests which were failing in his branch. Fixes #881 and #555.

- Ensure set's are encoded as lists
- Fixup to_string on the iterable fields to explicitly convert ints instead of relying on `str(list(...))`to avoid the trailing L's 
- Retain list order when deserializing list fields
- Change serialization tests to perform an end to end serialize+deserialize.

PR checklist:
- [x] Updated CHANGELOG.md 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
  - When running the local sandbox, if a port clash is detected then the next port will be used (this was broken before)
  - Accessing the Datastore from outside tests will no longer throw an error when using the test sandbox
  - The in-context cache is now reliably wiped when the testbed is initialized for each test
+ - The iterable set and list fields and their related fk variations didn't properly serialize / deserialize.
 
 ## v0.9.9 (release date: 27th March 2017)
 
--- a/djangae/fields/iterable.py
+++ b/djangae/fields/iterable.py
@@ -201,9 +201,15 @@ class IterableField(models.Field):
         # If possible, parse the string into the iterable
         if not hasattr(value, "__iter__"): # Allows list/set, not string
             if isinstance(value, basestring):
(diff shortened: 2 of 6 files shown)


nealedj:
I guess this is incompatible with https://github.com/potatolondon/djangae/pull/901.

This goes quite a lot further as it alters the current serialization behaviour to dump the iterable fields as valid JSON lists rather than encoding them as strings.

Close off #901 if this gets merged.

grzes:
Waiting for some fixes on 901 if we can make it work we're definitely grabbing 901 over this

Kazade:
Superseding with #901

PR #917, opened 2017-05-17
Ensure that the ordering of values in a RelatedListField is respected…

dannymilsom:
… via form update

- Previously if you were to try and set some values in the `RelatedSetField` via a form save, the ordering would not always be retained (if it were, this was by chance due to the underlying internals).

PR checklist:
- [ ] Updated CHANGELOG.md 
- [X] Added tests for my change


adamalton:
I've made a couple of minor tweaks to this in #922 if you fancy doing a quick review @dannymilsom

PR #930, opened 2017-06-04
Feature/app for production testing

### LINKED ISSUE #713: Performance of .count() seems slow
I've noticed that running .count() can be much MUCH slower than running len(things.all()). This ticket is to write some performance tests of both and see if that is the case, and if so, why!?

seawolf42:
Fixes #713.

- Add testprodapp as an app that can be deployed to GAE in production for testing against the live GCP environment.

PR checklist:
- [x] Updated CHANGELOG.md 

I am still working on documentation, this is just an early PR so that you can start to review.


seawolf42:
I've added some basic documentation to the `testprodapp` directory itself, let me know if you want references to this in the `docs` folder as well.

seawolf42:
Also, I currently own the djangae-testprodapp project in GAE, but I'm happy to transfer ownership and/or give others access. I currently have it set for a $0/day billing cap, so the only limitation would be extensive testing and/or large entity sets would not be possible while I'm the owner.

Kazade:
Thanks for this! I'll give it a proper review and merge when I can :)

PR #941, opened 2017-06-13
Add djangae.contrib.backup app

nealedj:
Adds a new contrib app to allow automatic backups to be configured.

This is basically a request handler which can be hit by a cron in order to call the datastore backup job provided by datastore admin.

The handler allows you to:

- Automatically backup all registered models without the need to manually update the backup URL when new models are added
- Allow the backup name and GCS bucket to vary based on the value in settings. For example, you may have a single codebase deployed to multiple apps (such as a staging environment) and want to back them all up to different GCS buckets or give them different names. Without this handler a backup URL would need to be static and I don't it'd be possible to back up any more than one app without manually altering the cron entry before deploying to each app.

PR checklist:
- [x] Updated CHANGELOG.md 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
   This will be removed in the next release of Djangae so you'll need to re-save your entities with this setting set to False before upgrading (see [detailed release notes](release_notes/0_9_10.md)).
 - Added support for the 1.9.54 SDK
 - Implemented a full application that can be deployed to production GAE for testing real-world scenarios against GCP environment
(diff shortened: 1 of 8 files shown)


Kazade:
This looks really cool, just a few suggestions/questions:

1. Could we change the settings prefix from `DS_BACKUP` to `DJANGAE_BACKUP` for consistency (so all Djangae apps namespace their settings under DJANGAE)?
2. Could we rename get_datastore_setting to get_backup_setting?
3. Is there a reason why the example queue has a rate of only 1/s ?

PR #944, opened 2017-07-05
Core paginator fixes

### LINKED ISSUE #839: Optimize/fix djangae.core.paginator
The intention of this class was to *not* count or evaluate the entire queryset. It definitely seems to!

Kazade:
Fixes #839 

- Page queries are now executed when .page() is called
- Fixed bugs where the original queryset was manipulated when page() was called
- Wrote some tests

PR checklist:
- [] Updated relevant documentation
- [x] Updated CHANGELOG.md 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 
 ### Bug fixes:
 
- -
+ - Fixed performance issues and bugs in the Djangae core paginator
 
 ## v0.9.10
 
--- a/djangae/core/paginator.py
+++ b/djangae/core/paginator.py
@@ -16,7 +16,7 @@ class DatastorePaginator(object):
 
     def __init__(self, object_list, per_page, orphans=0,
                  allow_empty_first_page=True):
-        self.fetched_objects = object_list
+        self.object_list = object_list
         self.per_page = int(per_page)
         self.allow_empty_first_page = allow_empty_first_page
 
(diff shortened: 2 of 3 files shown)


adamalton:
Nice!

2 minor suggestions (but which should make it much clearer what's happening):

1. Rename some variables.

```python
class DatastorePaginator(object):

    def __init__(self, object_list, per_page, orphans=0, allow_empty_first_page=True):
        self.fetched_objects = object_list
        ...
```

* `object_list` is probably (or definitely?) a queryset.
* `self.fetched_objects` is not fetched at this point!

I suggest that we change both of those to `queryset`, or if we support using the paginator with a plain list instead of a queryset, then call them both `objects`.

Note that the `object_list` variable/attribute in the `DatastorePage` class *is* a list, so that should be left as it is.

2. Remove the handling of querysets from `DatastorePage`, because we never pass it a queryset.

```python
class DatastorePage(collections.Sequence):
    ...
    def __getitem__(self, index):
        ...
        # The object_list is converted to a list so that if it was a QuerySet
        # it won't be a database hit per __getitem__.
        if not isinstance(self.object_list, list):
            self.object_list = list(self.object_list)
```

I think we can just remove that comment and `if` block.

Agree?

Kazade:
So object_list is the argument name because of this: https://docs.djangoproject.com/en/1.10/topics/pagination/#paginator-objects

It's also the property on the Page object for the same reason. But yeah, I agree with the other things :)

PR #945, opened 2017-07-07
Require certificate validation for Python libraries creating HTTPS requests

moarso:
The Google App Engine runtime maintains backwards compatibility by not enforcing certificate validation checks for SSL/TLS for HTTPS requests (and others) despite this being the default behaviour in recent versions of Python. New applications should be able to require certificate validation without introducing any issues and thus improve transport layer security.

Using OpenSSL
https://cloud.google.com/appengine/docs/standard/python/sockets/ssl_support

PEP 493 -- HTTPS verification migration tools for Python 2.7
https://www.python.org/dev/peps/pep-0493/


moggers87:
You've only changed the app.yaml files for the test apps - this won't have any affect in any other app using Djangae. I suggest you close this PR and open a new one for https://github.com/potatolondon/djangae-scaffold

moarso:
Thanks! I'll create a new PR for djangae-scaffold. Did you want me to close this PR or does it make sense to commit this to the example apps as well?

moggers87:
Djangae-scaffold is the example/template app, I doubt ordinary users will be looking at the test apps.

Kazade:
Closing this in favour of a PR to scaffold

PR #946, opened 2017-07-13
Improve sql representation

### LINKED ISSUE #942: connection.queries doesn't show LIMIT being applied
When using a model backed by the nosql backend, changelist views within Django admin do not have a `LIMIT` clause on them, meaning when you get a lot of nosql model instances, load time increases exponentially for the admin views, eventually making them unusable.

This is because django does a `COUNT * FROM table` before determining whether to use the LIMIT clause, perhaps some sort of an optional admin mixin might help to resolve this?

..Or implementing a count method on the db backend that does smth like https://stackoverflow.com/a/4235169 potentially with some sharded caching on insert/delete..

Kazade:
Fixes #942 

- Greatly improves the approximate SQL representation in connection.queries

PR checklist:
- [x] Updated CHANGELOG.md 


adamalton:
Did you mean to revert that commit?

igniteflow:
My mistake, I've fixed this now on master

adamalton:
👍

PR #951, opened 2017-08-07
gauth backend fails to authenticate

farridav:
Within the auth backends, the atomic function/decorator/context manager is stored as a class property (to abstract it for use with Django/Djangae implementation), which when called [here](https://github.com/potatolondon/djangae/blob/master/djangae/contrib/gauth/backends.py#L139) , is treated as a class method, and a class instance is passed as the first argument.

This causes it to fail, as the first argument is expected to be the db alias (`using`), to fix this, we could either return the atomic function from another function, or, as i have done here, wrapped it in `partial`, so it is treated as a standalone function, not a classmethod.

Also included @TysonRV 's import fix

### DIFF
--- a/djangae/contrib/gauth/backends.py
+++ b/djangae/contrib/gauth/backends.py
@@ -1,4 +1,5 @@
 import warnings
+from functools import partial
 
 from django.db.models.query_utils import Q
 from django.db.utils import IntegrityError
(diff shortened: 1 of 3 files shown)


davidwtbuxton:
Nice bug!

Kazade:
Clever fix, thanks!

PR #957, opened 2017-08-14
System check errors should have an ID

moggers87:
- Add IDs to system check errors
- Add tests for checks
- Fixed check that didn't always return a list

Without IDs on errors raised by Djangae checks there is no way to silence them. A few examples:

* The check for django-sessions-csrf is not required with Django 1.11 as the same functionality is now built-in
* The check for unsafe-* CSP directives should be silenced if you want to use CSP nonces, but also want to support IE 11
  * To do this you need both nonces and `'unsafe-inline'` in your CSP directives, the latter being ignored by Firefox and Chrome when there is a nonce present

PR checklist:
- _n/a_ Updated relevant documentation
- [x] Updated CHANGELOG.md 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Add support for querying JSONFields in a similar way to the PostgreSQL JSONField
 - Allow special indexers to index `None` as well as remove unused index properties from the entity
+- Added IDs to system check errors, allowing them to be silenced
 
 ### Bug fixes:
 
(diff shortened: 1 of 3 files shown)


adamalton:
Have you run the tests?  `test_template_loader_present` looks like it's checking for the wrong error IDs (it's checking for the IDs of the CSP errors), and it's checking using `err.idi`, which is presumably a typo?

moggers87:
Yup, and so did Travis. That for loop is just in the wrong test method!

I'll push a fix in just a moment.

PR #962, opened 2017-09-03
Fix adding to Related Set/List Fields via reverse relations

### LINKED ISSUE #961: Reverse relationships do not work for RelatedSetFields.
For a quick example, create a group and then add a user to that group with the following snippet.

`group.gaedatastoreuser_set.add(user)`

This will throw this error:

`AttributeError: 'Group' object has no attribute 'groups_ids'`

It appears that the reverse relationship is not identifying itself correctly and django spazzes out. If you'll see it's being misidentified:

```
>>> group.gaedatastoreuser_set.field.attname
'groups_ids'
```

I'm unsure of a workaround at this time.

moggers87:
Fixes #961

- Fix adding and removing items from related set/list fields via reverse relations
- Factor out type checks from `RelatedIteratorManagerBase.add` and do these checks on `remove` too
- Additional tests for type checks that appeared to be missing

PR checklist:
- (N/A) Updated relevant documentation
- [x] Updated CHANGELOG.md 


adamalton:
I'm not sure that we should support this functionality.  See my comments on #961.

moggers87:
OK, I'll close this PR and open a new one.

PR #967, opened 2017-09-25
Fix runserver port

### LINKED ISSUE #948: manage.py runserver host:port broken on 0.9.10
With 0.9.10, `runserver` command with custom port (and host) has broken.

E.g. `./manage.py runserver 0.0.0.0:7350`

```
System check identified no issues (0 silenced).
July 24, 2017 - 12:14:45
Django version 1.10, using settings 'settings'
Starting development server at http://0.0.0.0:7350/
Quit the server with CONTROL-C.

INFO     2017-07-24 12:14:45,650 devappserver2.py:116] Skipping SDK update check.
INFO     2017-07-24 12:14:45,652 api_server.py:313] Starting API server at: http://0.0.0.0:8010
INFO     2017-07-24 12:14:45,655 dispatcher.py:226] Starting module "default" running at: http://0.0.0.0:8000
INFO     2017-07-24 12:14:45,661 admin_server.py:116] Starting admin server at: http://0.0.0.0:8011
```

The ports aren't configured properly anymore.

@Kazade I found this commit: 5d20e7c981af6f9b737517f70820bf776e86ab3d where you remove the port configuration in `djangae/management/commands/runserver.py`, is there a reason for that?
If you need to keep the default port values as constants in `djangae/sandbox.py`, maybe we could have the configuration back in `runserver.py`  that would override the constants?

Thanks!

igniteflow:
Fixes #948

- Fix passing of port (and optionally host) from runserver to devappserver

PR checklist:
- [x] Updated CHANGELOG.md 


adamalton:
LGTM.  I've also tried it out locally and it works.  Nice!

Had you considered just calling `super()` and getting [Django's code](https://github.com/django/django/blob/stable/1.11.x/django/core/management/commands/runserver.py#L70-L101) to do the parsing?  If we do that, then we call `super()`, Django's code will call `self.run()` for us, but we could then just check `self.addr` and `self.port` in the top of our `run()` method and update `self.gae_options` accordingly.

That way we'd support all of the [crazy magic that Django supports](https://docs.djangoproject.com/en/1.11/ref/django-admin/#examples-of-using-different-ports-and-addresses) (or as much of it as the SDK supports).

Thoughts?

adamalton:
Nice.  Although...

```TypeError: '8006' has type <type 'str'>, but expected one of: (<type 'int'>, <type 'long'>)```

It needs to `int()` the port number :-)

igniteflow:
Wat?  Argh I fixed this on my local app version.  Really should just set up with `pip install -e` instead of manually copying over

adamalton:
That's the badger!  Nice work!

PR #977, opened 2017-10-02
Replace print with logging

### LINKED ISSUE #976: Silence output during `manage.py test` execution
Unit tests and other functions are using print statements for output. This should be changed by:
* turning all `print()` calls in unit tests to `logging.XXX()` calls
* silencing all `print()` statements in command and other output when `test` is in `sys.argv`

seawolf42:
Fixes #976 .

- turn all `print()` calls in unit tests to `logging.debug()` calls
- silence all `print()` statements in command and other output when `test` is in `sys.argv`

PR checklist:
- [N/A] Updated relevant documentation
- [X] Updated CHANGELOG.md 
- [N/A] Added tests for my change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@
  - Fixed performance issues and bugs in the Djangae core paginator
  - Fix several issues with the test sandbox
  - Initialize the app_identity stub in the test sandbox
+ - Replace `print()` statements with `logging.debug()` in all unittests
+ - Silence stdout output during testing
 
 ## v0.9.10
 
--- a/djangae/contrib/security/tests.py
+++ b/djangae/contrib/security/tests.py
@@ -1,10 +1,12 @@
+import logging
+
 from djangae.test import TestCase
 from djangae.contrib.security.management.commands import dumpurls
 
 
 class DumpUrlsTests(TestCase):
     def test_dumpurls(self):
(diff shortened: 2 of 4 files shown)


adamalton:
This looks good.  But what is baseline.out?  Looks like it's been added by mistake?

seawolf42:
Yup, added by mistake. Removed from the history and latest changes from master are merged, ready to be looked at again when you have a chance, @adamalton. Thanks!

adamalton:
Great, thanks @seawolf42.  One minor thing, which is quite pedantic but would be nice to do, is to change the repeated occurrences of

```python
if not IN_TESTING:
   print (...)
```

with a method on the class.  I.e.:

```python
class DjangaeMigration(object):

    def print(self, value):
        if IN_TESTING:
            print(value)
```

and then use `self.print(...)` in place of `print(...)`.  It would just reduce the repetition, would allow us to easily change the logic of how/whether we print things, and would also allow subclasses to use the same logic.

Sorry, I should have mentioned this the first time! 😬

seawolf42:
That's a great idea, and I'm a little embarrassed I didn't think of it when I was writing this, I was cringing at the code smell but it didn't occur to me to do it that way. I will submit an updated PR in a few minutes.

adamalton:
Nice.  Thanks!

PR #979, opened 2017-10-11
Add github issue template.

adamalton:
This will hopefully help avoid so many questoins being submitted as issues.

What do you think @Kazade?


WillsB3:
Maybe link to this in the main README too?

adamalton:
@WillsB3 the Google Groups group is already linked to in the README.  Is that what you were meaning?

PR #1000, opened 2017-11-13
Switching the import order, as in sdk 1.9.54 simply attempting to imp…

### LINKED ISSUE #964: Breaking change (from SDK), sandbox moved in 1.9.60
Having updated the SDK to the latest build, alot of classes have moved inside a `runtime` package, to resolve the issue I have replaced `google.appengine.tools.devappserver2.python`  with `google.appengine.tools.devappserver2.python.runtime` in all of the Djangae imports

grzes:
…ort something from devappserver2 here results in an attribute error when devappserver tries to import mail_stub.

Fixes #964 (again)

### DIFF
--- a/djangae/compat.py
+++ b/djangae/compat.py
@@ -9,7 +9,7 @@ except ImportError:
     from google.appengine.tools.devappserver2.python.runtime import runtime
 
 try:
-    from google.appengine.tools.devappserver2.devappserver2 import _LOG_LEVEL_TO_RUNTIME_CONSTANT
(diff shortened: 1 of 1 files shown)


igniteflow:
@grzes what import were you attempting to make that was failing?  I don't see how this order change fixes anything.  With `1.9.54` installed I can import `_LOG_LEVEL_TO_RUNTIME_CONSTANT` successfully on both `master` and `compat-fix`.

PR #1006, opened 2017-11-29
Changed djangae.contrib.backup to be easier to use.

### LINKED ISSUE #1004: The djangae.contrib.backup app is a pain to use
2 things annoy me when using the djangae.contrib.backup application.

First is the need to configure the bucket name (which should default to the application's default bucket name) and a name for the backup (which should default to something sensible like "djangae-backups"). It should be enough to just set `DJANGAE_BACKUP_ENABLED = True` and you're done. Those projects that need to fine tune the settings can continue to do so.

Second is that the backups are always created in the same directory in the bucket. Because the Google datastore backup format is somewhat crazy (a bunch of files and directories with totally meaningless random names) this means you quickly have a backup directory full of hundreds (if not thousands) of objects, and identifying the most recent backup, or what backups have been created, is tedious from the Google Cloud Storage browser.

What would be much nicer is if the backups were created with a new directory each time, with the new directory name including a timestamp. Something like "myapp.appspot.com/djangae-backups/20171127-235900/". That way it would be easy to tell when backups were created, etc.

Third, the documentation is out of date.

Can I fix?

Dave

davidwtbuxton:
Fixes #1004.

- The DJANGAE_BACKUP_GCS_BUCKET setting is optional. It defaults to
    the project's default bucket with a suffix of 'djangae-backups',
    e.g. 'myapp.appspot.com/djangae-backups'.
- The DJANGAE_BACKUP_NAME setting is optional. It defaults to
    'djangae-backups'.
- Backups are created with a timestamp in the object name, e.g.
    'myapp.appspot.com/djangae-backups/20171128-235900'.
- Updated docs (and fixed the docs config to include the backup docs).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### New features & improvements:
 
+- Backups made with `djangae.contrib.backup` are created in a new, time-stamped directory to make managing backups easier. Only `DJANGAE_BACKUP_ENABLED` is required, all other backup settings are optional and the default is to create backups in the default cloud storage bucket. See the backup docs for details.
 - Add support for querying JSONFields in a similar way to the PostgreSQL JSONField
 - Allow special indexers to index `None` as well as remove unused index properties from the entity
 - Added IDs to system check errors, allowing them to be silenced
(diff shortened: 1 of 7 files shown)


Kazade:
Great work!

PR #1019, opened 2017-12-18
Use '' instead of None for an unset namespace.

rspier:
Fixes broken memcache caching if the default DATABASE namespace is set to ''.

This matches the default namespace returned by App Engine.  Otherwise, you
end up in a situation where the getter code uses a (memcache) key with a stringified
None, but the setter code is using a stringified ''.  i.e.  You can't
retrieve anything from the cache.

An alternate fix would be to fix the add_... code to do the same or None check

- Changes default namespace used by memcache to '' instead of None

This change takes our cache hit rate from ~8% up to >90% in testing.  I don't have historical cache hit data, but I suspect this was made worse by the [workaround](https://github.com/potatolondon/djangae/issues/789#issuecomment-264225096) for issue #789

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
  - It was possible a `TypeError` would throw when calculating the ComputedCollationField value if the source value was unicode
  - Make `value_from_datadict` in `forms.fields.ListWidget` return None when the value provided is None as the existing comment describes. This prevents an exception when `save()` is called on a `ListWidget` whose value is `None`.
  - Fixed test to remove dependency on mock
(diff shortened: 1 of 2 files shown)


rspier:
@nealedj FYI

Kazade:
Nice one! Thanks!

nealedj:
For reference, it appears that this will only have broken this caching behaviour if the namespace of the database was explicitly set to `""`.

I.e. `DATABASES['default']['NAMESPACE'] = ""`

rspier:
Thanks!

PR #1024, opened 2018-01-24
Security tweaks

Kazade:
- Makes the CSRF session check respect the newish `CSRF_USE_SESSIONS` setting in Django
- Locks down the clearsessions view to tasks or admins only

PR checklist:
- [x] Updated CHANGELOG.md 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Added custom `FileField` and `ImageField` which accept an optional `url_field` argument to allow you to specify the name of another field on the model on which the URL to the file or image is cached.
 - Add a ComputedNullBooleanField
 - Updated the `sleuth` library in djangae.contrib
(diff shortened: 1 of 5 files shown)


adamalton:
You need to remove this from the `clearsessions` view:

```python
if not environment.is_in_cron():
         return HttpResponse(status=403)
```

adamalton:
I'm wondering whether we should only check for the `CSRF_USE_SESSIONS` setting if the Django version is new enough to respect that setting.  Otherwise you could be on Django 1.8 and set `CSRF_USE_SESSIONS` to `True` and the check would incorrectly assume that you're using session-based CSRF.  Thoughts?

Kazade:
@adamalton look OK to you?

adamalton:
Yep, was just waiting for the tests to finish.

PR #1029, opened 2018-02-13
Tweak retry time

Kazade:
- This increases the min wait time for retry from 100ms to 375ms as the previous value caused more problems than it solved (increased contention and likelihood of transaction failure)
- The value is now overridable through the `_initial_wait` kwarg.

This is a stop-gap improvement as `retry()` needs a proper cleanup and tests. I haven't added documentation as the function is currently undocumented and will be undergoing API change soon so it's probably best not to advertise it too much now!

PR checklist:
- [x] Updated CHANGELOG.md 


adamalton:
```python
time.sleep(timeout_ms / 1000000.0)
```

I think we've got too many zeros in there.  `375 / 1000000` is `0.000375`.  That's not a very long wait time!

Maybe that's the source of your problems?

PR #1036, opened 2018-02-27
Fix cross-database support for RelatedSetField and RelatedListField

### LINKED ISSUE #1034: `RelatedSetField` and `RelatedListField` related querysets are always empty if the related model uses a different database
`RelatedSetField` and `RelatedListField` related querysets are always empty if the related model uses a different database. This is because `RelatedIteratorManagerBase` uses the instance model rather than the related model when routing databases, resulting in the wrong database being queried.

* Djangae version: 0.9.9
* Django version: 1.9
* App Engine SDK version: 1.9.38
* Using Cloud SDK: no

RoganMurley:
Fixes #1034.

- Fix cross-database support for `RelatedSetField` and `RelatedListField`.
- Introduce `testapp.routers.TestRouter` so that models in tests can explicitly choose a database.

PR checklist:
- [x] Updated CHANGELOG.md 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@
  - Restricted access to the `clearsessions` view to tasks and admins only
  - Fixed the `sleep()` time in `djangae.utils.retry` which was sleeping in `ns` rather than `ms`
  - Fix unicode error when creating a SQL representation
+ - Fix cross-database relationship support for `RelatedSetField` and `RelatedListField`.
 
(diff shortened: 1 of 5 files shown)


adamalton:
Given that `ForeignKey` doesn't work across databases, I think we should explicitly state in the documentation that it's supported for `RelatedXField`s.  Otherwise, no one is ever going to guess that it's a thing.

PR #1039, opened 2018-02-28
Lock down the backup view

### LINKED ISSUE #1038: Backup view isn't protected
Issue description

If you include the Djangae contrib backups up, the create_datastore_backup view has no protection. It should be locked down to tasks or admins only using the decorator in djangae.environment.

The URL is also missing a trailing slash, but we'll have to make that optional as people will be using it already.

Steps to reproduce

Install the `djangae.contrib.backups` app, run `./manage.py dumpurls`

Traceback or other info

* Djangae version: 0.9/10
* Django version: N/A

Kazade:
Fixes #1038 

- Locks down the backup creation view to admins and tasks
- Fixes the URL to match the documentation

PR checklist:
- [x] Updated CHANGELOG.md 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,8 @@
  - Fixed the `sleep()` time in `djangae.utils.retry` which was sleeping in `ns` rather than `ms`
  - Fix unicode error when creating a SQL representation
  - Fix cross-database relationship support for `RelatedSetField` and `RelatedListField`.
+ - Locked down the backup creation view in `djangae.contrib.backups`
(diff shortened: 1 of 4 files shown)


adamalton:
Does the `/?` at the end of the URL regex definitely work?  I seem to recall trying that and it not working in the past.

Kazade:
Yep it works: https://stackoverflow.com/questions/1596552/django-urls-without-a-trailing-slash-do-not-redirect#11690144

It'll bypass any Django APPEND_SLASH redirect behaviour, but as this is designed to be a cron task that won't work anyway as cron tasks don't follow redirects.

adamalton:
👍

PR #1049, opened 2018-04-09
Fix computed collation field

### LINKED ISSUE #1048: ComputedCollationField fails in multiple ways
Issue description

ComputedCollationField will throw an error like this if you are computing the collation from a long string.

```
BadRequestError: The value of property "XXXXXX" is longer than 1500 bytes.
```

This is because the collation string is much longer than the original. A better behaviour here would be to log a warning and truncate the data. As this field is used for ordering it's unlikely that differences beyond the first 1500 bytes will make a huge difference to ordering. It's far better than dying on save.

Steps to reproduce

Traceback or other info

* Djangae version: 0.9.10
* Django version: 1.11
* Using Cloud SDK: no

Kazade:
Fixes #1048 

- Fixes incorrect sort key generation
- Warns and truncates sort keys which are too long

PR checklist:
- [x] Updated CHANGELOG.md 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,7 +47,7 @@
  - Fix cross-database relationship support for `RelatedSetField` and `RelatedListField`.
  - Locked down the backup creation view in `djangae.contrib.backups`
  - Fixed the backup creation URL to have a trailing slash (optional, to prevent breaking apps)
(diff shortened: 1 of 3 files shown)


adamalton:
Nice.

We should probably make it clearer in the changelog that there was a bug and that existing data will be broken.

Also, when we log the truncation, it should log the model as well as the field.

PR #1054, opened 2018-05-18
Improve Python 3 Compatibility

dannymilsom:
Attempt to make our codebase better support both Python 2 and 3 simultaneously.

Note this is very much a work in progress and does not promise to provide full support for Python 3 yet - but could be merged into master as we work towards this goal incrementally. For this reason I have not updated the documentation.

- Simultaneously support for Python 2 and 3

PR checklist:
- [ ] Updated CHANGELOG.md 

### DIFF
--- a/djangae/storage.py
+++ b/djangae/storage.py
@@ -1,5 +1,4 @@
 # coding: utf-8
-import urllib
 import mimetypes
 import re
 import threading
@@ -335,7 +334,7 @@ class CloudStorage(Storage, BlobstoreUploadMixin):
 
             # Also, Django requires that url() return something 'truthy' even if the file field hasn't been
             # saved yet so we do the same thing if the file is not found (just add the bucket to the filename)
-            quoted_filename = urllib.quote(self._add_bucket(filename))
+            quoted_filename = six.moves.urllib.parse.quote(self._add_bucket(filename))
             return '{0}{1}'.format(self.api_url, quoted_filename)
 
     def _get_blobkey(self, name):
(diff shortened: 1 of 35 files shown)


davidwtbuxton:
I think all of the places where we use `dict.iteritems()` and `xrange()` should be replaced with `dict.items()` and `range()` (or for dictionaries, iterating over the dictionary directly where that makes sense). We shouldn't use the six shim for these.

`dict.iteritems()` and `xrange()` are useful for performance optimisations (memory pressure). We don't need those optimisations. Avoiding the use of six altogether would be better.

djangae/storage.py
```diff
@@ -7,7 +7,7 @@
 try:
     from cStringIO import StringIO
 except ImportError:
-    from StringIO import StringIO
+    from io import StringIO
```
davidwtbuxton: `StringIO` isn't used in this module. We can remove it and `re`.

adamalton:
Just one Danny, this looks good!

I'm indifferent about the six or not-six approach to `range` and `dict.iteritems`.  I guess using less `six` would be good, but given that we're using it anyway...?  I'll let you decide!

Other than that, there's just one failing test.

davidwtbuxton:
> I'm indifferent about the six or not-six approach to range and dict.iteritems.

Not relying on six where possible reduces complexity and will be better for the code in the long run. None of the places where the code is using `dict.iteritems()` needs to be concerned with memory use, and should have made use of `dict.items()` in the first place.

Of the code that makes use of `xrange()`, only in the deferred task stuff does it look like we may be building a sequence of potentially hundreds of items, and even in that case a list of 1,000 integers occupies less than 10 kilobytes.

Six should be used only when there isn't a more idiomatic syntax available. If it is possible to write code that works on Python 2 and 3 as-is, then we should write code that works on Python 2 and 3 as-is.

adamalton:
Fair.

PR #1062, opened 2018-05-31
Changed the logic for generating keys when sharding.

### LINKED ISSUE #1057: ValueError in _get_range function
### Issue description
Calling `djangae.db.migrations.mapper_library._get_range` with keys `60a1fef9136a4584a1cbf8a3193394b6` and `5e840832631344adb297493a5aade1bc` raises a `ValueError: unichr() arg not in range(0x110000) (wide Python build)` error.

### Steps to reproduce
[See test](https://github.com/andreipetre/djangae/blob/get_range-valueerror/djangae/tests/test_migrations.py#L689).

Traceback or other info

* Djangae version: master@815523c1a1352042b38fa019843452175c021046
* Django version: 1.8.15
* App Engine SDK version: 1.9.57
* Using Cloud SDK: no

### LINKED ISSUE #1060: Migrations skips entities
Hi,

I think there is a bug with Djangae's migrations framework where it will always skip over entities which use a numeric ID and the ID is either very small or very large. **This means that the migrations will not touch all entities.**

I believe migrations works roughly as follows:

- [Call `mapper_library.start_mapping()`][1].
- `mapper_library.start_mapping()` [calls `mapper_library.shard_query()`][2] to get a list of `(start, stop)` key pairs.
- Create a `ShardedTaskManager` entity and [save the list of pairs as `shards_queued`][3].
- `ShardedTaskManager.begin_processing` is [called in a deferred task][4].
- For each `(start, stop)` key pair, `ShardedTaskManager.begin_processing` [fires `ShardedTaskManager.run_shard`][5].
- `ShardedTaskManager.run_shard` queries for entities [filtered by `__key__ >= start` and `__key__ < stop`][6] and applies the desired migration function to each entity.  

The big bug here is in the logic to generate the list of `(start, stop)` key ID pairs that are later used to iterate over all entities. The `shard_query()` function does the following:

- Find the lowest value entity key ID, called `min_id`.
- Find the highest value entity key ID, called `max_id`.
- Use a `__scatter__` query order to get a sample of valid entity keys.
- Add the `min_id` key at the start of the list of valid keys.
- Add the `max_id` key at the end of the list of valid keys.
- Call `_generate_shards()` to get an initial list of `(start, stop)` key pairs.
- More logic to insert additional key pairs until we have at least the desired number of shard pairs (most of the time).

But the `_generate_shards()` function is broken. It is supposed to return a list of pairs where the first item of the first pair is equal to the min_id, and the second item of the last pair is equal to the max_id. This ensures that the later filtering and iterating over entities will include all entities.

Instead the function drops the first and last pair in the sequence of pairs it needs to make.

Quick testing in the shell:

```
>>> from djangae.db.migrations.mapper_library import _generate_shards
>>> keys = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
>>> _generate_shards(keys, 1)
[[1, 10]]
>>> _generate_shards(keys, 2)
[]
>>> _generate_shards(keys, 3)
[[4, 8]]
>>> _generate_shards(keys, 4)
[[4, 6], [6, 9]]
>>> _generate_shards(keys, 5)
[[3, 5], [5, 7], [7, 9]]
>>> _generate_shards(keys, 10)
[[1, 2], [2, 3], [3, 4], [4, 5], [5, 6], [6, 7], [7, 8], [8, 9], [9, 10]]
```

In the test, the only time it returns the correct result is when shard_count is 1 or is equal to the number of keys.

I think this probably isn't too difficult to fix, and once fixed a lot of the code to insert extra shards can also be removed.

[1]: https://github.com/potatolondon/djangae/blob/v0.9.11/djangae/db/migrations/operations.py#L98
[2]: https://github.com/potatolondon/djangae/blob/v0.9.11/djangae/db/migrations/mapper_library.py#L442
[3]: https://github.com/potatolondon/djangae/blob/v0.9.11/djangae/db/migrations/mapper_library.py#L459
[4]: https://github.com/potatolondon/djangae/blob/v0.9.11/djangae/db/migrations/mapper_library.py#L467
[5]: https://github.com/potatolondon/djangae/blob/v0.9.11/djangae/db/migrations/mapper_library.py#L403
[6]: https://github.com/potatolondon/djangae/blob/v0.9.11/djangae/db/migrations/mapper_library.py#L341

Tested with:

- Djangae f6aff6de04bbc8e9e049e4c9cd8f850fac24bced (current head of master).
- Django 1.11.10.
- App Engine SDK 1.9.57.
- Not using Cloud SDK.

davidwtbuxton:
We had a bug where keys that were very high or very low in the range of
valid keys would be ignored when iterating over entities (so the data
migration was not being applied for all entities).

The fix is to make sure that we always start one shard with the very
lowest key ID, and that the last shard has an unbounded upper range
for keys.

Needs more tests.

Fixes #1060.

These changes simplify the logic of choosing entity keys to use when filtering entities in a shard. I removed the logic that tries to calculate a key value half way between 2 known keys, which means this also fixes #1057 (because the code no longer deals directly with key values).

I tested this on App Engine with a few hundred entities, where the keys used integer IDs. I have also tested this using dev_appserver.py with a few thousand entities, for keys which use integer IDs and which use string IDs.


kevincarrogan:
It looks like the tradeoff here is that in fixing this bug the number of shards may be fewer than the number of shards desired but this fix, most importantly, means that no entities are missed off (which may happen with the current implementation).

What is the reason for wanting a particular number of shards? If it's to limit to a maximum desired amount and no more, then it somewhat seems reasonable that this fix may end up producing fewer shards than expected (with the benefit of not missing off entities).

If the desire is that we want the most amount of shards possible up to a certain amount, and getting the most we can is important, then I'm presuming this fix is less than desirable.

adamalton:
If my understanding is correct, then the only time when we would get fewer shards than specified is when the investigative query using the `__scatter__` property doesn't return enough entities.  But the more entities you have the less likely that is to happen.  And therefore the chance of not having as many shards as requested will decrease as the necessity for those shards increases.  Therefore I think it's fine to remove the logic for doing further splitting of shards. https://github.com/GoogleCloudPlatform/appengine-mapreduce/wiki/ScatterPropertyImplementation

My only question is: why do we remove the upper limit but not the lower limit?  I.e. why do we set the last shard to `(<key>, None)` but we don't set the first shard to `(None, <key>)`?

davidwtbuxton:
> My only question is: why do we remove the upper limit but not the lower limit? I.e. why do we set the last shard to (<key>, None) but we don't set the first shard to (None, <key>)?

Because that was closest to what the previous logic was doing. It sounds like it would work to have the first query shard have no lower limit, and I think I see what the change would be to implement this.

I also liked the fact that if we leave the last range open-ended then we use an less-than filter all the time, instead of having to use less than or equal so that we include the final key. (But this is a slightly different point to whether the first pair of keys should start with `None`.)

adamalton:
In that case, I think it makes sense to make both "ends" of the sharding the same, and remove the lower limit from the first shard.

davidwtbuxton:
I'm reluctant to make the change to have the starting pair start with `None` because I haven't tested that solution, and I would want to go back and actually test that solution before submitting a new PR. The current logic has been tested and used and I'm confident it works.

PR #1099, opened 2018-08-28
Transaction should track created entities

### LINKED ISSUE #1098: transaction.has_already_been_read should account for created objects
When using `djangae.db.transaction.atomic`, if you create a new object and you then call `transaction.has_already_been_read(obj)`, it will refresh the object from the DB, but it won't exist because the transaction will cause the DB state to appear as it was at the start of the transaction.

With Djangae's caching switched on, this won't be a problem, because the caching will return the object anyway.  But without caching it will break.

Example:

```python
from djangae.db.transaction import atomic, current_transaction

@atomic()
def func():
    obj = MyModel.objects.create(pk=1)
    
    # This may look like an unlikely use case, but with nested transactions it could easily happen
    current_transaction().refresh_if_unread(obj)  # Will raise DoesNotExist
    obj.thing = 1
    obj.save()
```

This is all theoretical based on my reading of the code.

* Djangae version: master@169599a03cef59f0e0770adeb79438220a6cf44b
* Django version: 1.11
* App Engine SDK version: 1.9.74
* Using Cloud SDK: no

adamalton:
Fixes #1098

Changed the behaviour of `transaction.refresh_if_unread` to:

* Not refresh objects which have been created within the transaction (because they won't exist at the DB level until the transaction is committed).
* Not refresh existing objects which have been re-saved within the transaction... because... that's the easiest thing to do!  But I don't know if it's the _right_ thing to do?!

PR checklist:
- [x] Updated CHANGELOG.md 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 - Additional option to not start mapper pipeline; and provide outputs to finalize function.
 - `atomic()` (when used as a context manager) now returns an object representing the current transaction
 - Added `djangae.db.transaction.current_transaction()` to return the same thing from inside an `atomic()` decorator
-- Added `Transaction.has_been_read(instance)` and `Transaction.refresh_if_unread(instance)` which allows writing safe transactional code.
+- Added `Transaction.has_been_read(instance)`, `Transaction.has_been_written` and `Transaction.refresh_if_unread(instance)` which allows writing safe transactional code.
 
 ### Bug fixes:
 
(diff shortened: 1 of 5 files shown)


adamalton:
@Kazade I'm not sure whether I've created the right behaviour for `refresh_if_unread` when handling existing objects which have been re-saved within a transaction.  I've made it ignore them, but I'm not sure if that's the right thing to do.

Issue #1103, opened 2018-10-03
Production error


jacobg:
The following commit is causing startup failure in GAE production:
potatolondon/djangae@36974d0

Exception:
```
2018-09-27 20:16:35.804 EDT
    from djangae import checks
2018-09-27 20:16:35.804 EDT
  File "/base/data/home/apps/s~/1.412878905325365371/sitepackages/prod/djangae/checks.py", line 8, in <module>
2018-09-27 20:16:35.804 EDT
    from google.appengine.tools.sdk_update_checker import GetVersionObject, _VersionList
2018-09-27 20:16:35.804 EDT
ImportError: No module named sdk_update_checker
```

Looks like sdk_update_checker only exists in local dev sdk, not in production. And so the check should only be made locally, not in production.

Perhaps this is the fix in djangae/checks.py:
```
try:
    from google.appengine.tools.sdk_update_checker import GetVersionObject, _VersionList
except ImportError as e:
    GetVersionObject = None
```
```
@register
def check_app_engine_sdk_version(app_configs=None, **kwargs):
    errors = []
    if GetVersionObject:
        sdk_version = tuple(_VersionList(GetVersionObject()['release']))
        if sdk_version > MAX_APP_ENGINE_SDK_VERSION:
                errors.append(Warning(
                    "MAX_APP_ENGINE_SDK_VERSION",
                    hint="You are using a version of the App Engine SDK that is not yet supported",
                    id='djangae.W002',
                ))
    return errors
```

Do you agree this fix needs to be made?

davidwtbuxton:
Full traceback on app startup:

```

Traceback (most recent call last):
  File "/base/alloc/tmpfs/dynamic_runtimes/python27g/ec315266546cb44c/python27/python27_lib/versions/1/google/appengine/runtime/wsgi.py", line 240, in Handle
    handler = _config_handle.add_wsgi_middleware(self._LoadHandler())
  File "/base/alloc/tmpfs/dynamic_runtimes/python27g/ec315266546cb44c/python27/python27_lib/versions/1/google/appengine/runtime/wsgi.py", line 299, in _LoadHandler
    handler, path, err = LoadObject(self._handler)
  File "/base/alloc/tmpfs/dynamic_runtimes/python27g/ec315266546cb44c/python27/python27_lib/versions/1/google/appengine/runtime/wsgi.py", line 96, in LoadObject
    __import__(cumulative_path)
[...]

PR #1101, opened 2018-09-16
check google cloud sdk version on local server start

seawolf42:
Fixes #1100.

Summary of changes proposed in this Pull Request:
- add gcloud version check in `checks.py`
- tests

PR checklist:
- [ ] Updated relevant documentation
- [x] Updated CHANGELOG.md 
- [x] Added tests for my change

PR #1105, opened 2018-10-03
Fix/issue 1100

### LINKED ISSUE #1100: Google Cloud SDK version check
If the local version of the Google Cloud SDK is newer than the highest supported version, the user gets cryptic messages (usually related to import errors). Related to #1035, #1042, 

Steps to reproduce:
```bash
$ gcloud components install --version=171.0.0
$ python manage.py runserver
```

Traceback or other info:
(various messages, depending on which SDK version is installed)

* Djangae version: 0.9.11 (also master/HEAD)
* Django version: all
* App Engine SDK version: 1.9.58
* Using Cloud SDK: yes

seawolf42:
Fixes #1100

This is a re-submission of #1101 but with a fix for the production break reported as #1103.


seawolf42:
FYI the fix proposed by @jacobg in the comments in #1103 is exactly how I amended this, thanks for that!

adamalton:
I'm going to add a comment for why we're try/except-ing the import, but I'll just add that in master after merging.  Thanks for the fix!

PR #1109, opened 2018-11-07
Replace 'models.get_models()' with 'AppConfig.get_models()' in uniquetool app

marcoazza:
Fixes AttributeError when calling models.get_models(). Function get_models in models module has been deprecated and then removed in newer Django releases, it has been added as method of AppConfig class.

- Replace 'models.get_models()' with 'AppConfig.get_models()'


adamalton:
The fix looks good, thanks Marco.

I don't think that the `test_model_choices_provided_list` test is necessary though, as I think it's testing something which is not the intended behaviour.  The idea of the `_model_choices` attribute on the `UniqueActionAdmin` class is not to allow the list of models to be customised by subclassing it, it's just a caching thing so that the list of models which have unique fields doesn't need to be re-calculated each time the `model_choices()` method is called.

So I think you can just remove that test.  And then the docstring on the `test_model_choices_all_models` test needs updating - it currently says that `get_models()` should return all models, but actually it only returns models which have at least one field marked as `unique=True` or at least one entry in the `_meta.unique_together` definition.  I don't think it's worth changing the test to check that behaviour, but the docstring of the test currently implies a behaviour which isn't actually what it does.

And in the CHANGELOG entry, it should probably say "Fixed" instead of "Fixes" :-)

adamalton:
Nice.  You could still remove `When _model_choices is not defined` from that docstring, because the `_model_choices` attribute is actually irrelevant to the behaviour (it's just an implementation detail).

Also, any chance that you could merge `master` back into your branch, or rebase it on `master` to resolve the conflict in the CHANGELOG?

adamalton:
Nice one Marco!

PR #1110, opened 2018-11-08
Add `deferred.defer_iteration_with_finalize`

Kazade:
- Adds a nice method for iterating large querysets in deferred tasks

PR checklist:
- [x] Updated CHANGELOG.md 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@
 - Added `Transaction.has_been_read(instance)`, `Transaction.has_been_written` and `Transaction.refresh_if_unread(instance)` which allows writing safe transactional code.
 - Added `Transaction.has_already_been_read(instance)` and `Transaction.refresh_if_unread(instance)` which allows writing safe transactional code.
 - Added App Engine SDK version check on project startup.
-- Added support for named class-based views to dumpurls.  Also now supports export to either json or csv 
+- Added support for named class-based views to dumpurls.  Also now supports export to either json or csv
+- Added `deferred.defer_iteration_with_finalize`
 
 ### Bug fixes:
 
--- a/djangae/admin.py
+++ b/djangae/admin.py
@@ -1,0 +1,5 @@
+from django.contrib import admin
+
+from djangae.models import DeferIterationMarker
+
+admin.site.register(DeferIterationMarker)
--- a/djangae/db/migrations/mapper_library.py
+++ b/djangae/db/migrations/mapper_library.py
@@ -11,7 +11,6 @@ import itertools
 import logging
 from datetime import datetime
 
-from djangae.db.backends.appengine import rpc
 from django.conf import settings
 from django.utils.six.moves import range
 from google.appengine.api import datastore_errors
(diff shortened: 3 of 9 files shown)


adamalton:
I found the stuff in the documentation about it not handling `DeadlineExceededError` a bit confusing.  On first reading it made me think that I would need to add my own handling for it.  It makes sense now that I've read the code, but I think it could be worded better.  Maybe something like:

> Each processing task keeps track of its execution time and re-defers itself to avoid hitting App Engine's DeadlineExceededError.  However, this check is only performed in between the processing of each object and the re-deferring only happens when the task is within 15 seconds of hitting the deadline.  So if the processing of an individual model instance takes more than 15 seconds then the DeadlineExceededError may still be hit, which will cause that task to be retried from the beginning, thus re-processing some of the model instances.

Now that I've written that^ it makes me think that `_BUFFER_TIME` should be an optional parameter to `defer_iteration_with_finalize`.

Is there a reason why you can't pass `kwargs` for the callback function?  Seems like that's worth adding, right?

The transaction which does

```python
marker.shards_complete += 1
marker.save()
```

should probably be wrapped with `retry`.  Otherwise, if 2 shards finish at the same-ish time and get contention trying to update the marker (which is not unlikely) then one of them will retry that whole shard (or the last task of that shard) unnecessarily.

I think that `DeferIterationMarker` should have some kind of identifier on it.  Currently it doesn't even have a created date, which means that once you've used `defer_iteration_with_finalize` twice you have no idea which marker was which.  It probably wants some kind of textual identifier as well a timestamp though, such as `"%s on %s" % (callback.__name__, queryset.model.__name__)` or maybe an identifier that you can specify yourself as a parameter to `defer_iteration_with_finalize`.

Other than all of that stuff, it looks great :D

Also, I'm glad to see that I'm not the only person hitting that weird test failure with the `retry` tests in Travis!

jacobg:
Very cool. Is the purpose of this work to move this project forward to next-gen App Engine? It sounds like we both really just use Map-Reduce for mapping querysets, and don't really use the reduce part of it.

It would be useful to keep track of the number of entities that were iterated so it could be logged in the finalize function. Perhaps also store and update that # in DeferIterationMarker as the shards iterate, so it can be monitored. Maybe a more generic approach would be to allow the callback function to accept and return a value that gets chained through each successive callback, and then gets stored in DeferIterationMarker at the end of each slice.

I'm also in need of a replacement for pipelines. I don't want to have to setup Cloud Dataflow for some of the use cases I have. Do you also use pipelines directly?

Kazade:
@jacobg This isn't intended to replace all the functionality of mapreduce - but rather to deal with the most common use case directly in the core of Djangae without requiring MR as a dependency. You're right, this abstraction will be useful as we move to Python 3 and Google's Cloud Tasks etc.

I don't want to add any counting functionality, as counters (even sharded counters) introduce contention points and are very use-case specific. We've actually found that a more optimal way to deal with counting is to create a marker instance (each with a random group) on each `callback` and then count them in groups in the finalize part. If you wait long enough to ensure that you don't get bitten by eventual consistency this is fine (and moving forward with Datastore on Firestore, this won't be a concern).

We're trying to find time to come up with a proper roadmap for what Djangae 2 will look like (Djangae on Python 3) and it's going to be very different. Stay tuned :)

jacobg:
@Kazade Cool!

PR #1145, opened 2018-11-26
Test runner does skips import error tests

dannymilsom:
If you run the test suite with a target, and one of the targeted tests fails with an import error, this was not propagating to the test run output (it wasn't actually being run, so essentially failing silently).

This is because the test ID changes to have a prefix of `djangae.test_runner.ModuleImportFailure`.

You can verify this locally with the `testapp` by adding an invalid import to a test module, then running `manage.py test <path.to.the.test.module>`. Notice it will currently fail silently, as the test is ignored by the `DjangaeTestSuiteRunner`. 

PR checklist:
- [X] Updated CHANGELOG.md 

### DIFF
--- a/djangae/test_runner.py
+++ b/djangae/test_runner.py
@@ -212,7 +212,9 @@ class DjangaeTestSuiteRunner(DiscoverRunner):
             if hasattr(test, 'available_apps'):
                 test.available_apps = None
 
-            if args[0] and not any([test.id().startswith(x) for x in args[0]]):
+            # make sure we don't run any of the extra tests if a particular
+            # module target was provided at runtime
+            if args[0] and not any([x in test.id() for x in args[0]]):
                 continue
 
             if test.id() in DJANGO_TESTS_TO_SKIP:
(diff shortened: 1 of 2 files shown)


djangae/test_runner.py
```diff
@@ -213,8 +213,9 @@ def build_suite(self, *args, **kwargs):
                 test.available_apps = None
 
             # make sure we don't run any of the extra tests if a particular
-            # module target was provided at runtime
-            if args[0] and not any([x in test.id() for x in args[0]]):
+            # module target was provided at runtime. If there's an import error
+            # the test runner will add a colon separated prefix so we exclude that.
```
davidwtbuxton: > will add a colon separated prefix

Does this behaviour (specifically, using a ':' to separate the module name from test case name) depend on the runner? If it does, do we care?

I'm guessing that this may not be the case when the test suite runner is the default Django runner (which uses a '.' between the module name and the class name).

https://docs.djangoproject.com/en/1.11/topics/testing/overview/#running-tests

PR #1150, opened 2018-11-30
Improve retry

### LINKED ISSUE #1149: `retry()` doesn't catch InternalError and other improvements
Issue description

`djangae.utils.retry` can be improved in a few ways:

 - The default catch list doesn't include `datastore_errors.InternalError`
 - The min wait time is probably too low (most things need about a second to reliably retry but I think a 750ms min wait is probably OK)
 - There is no random element to the retry time which means several concurrent requests trying to update the same thing may all back-off and retry at the same time and then collide again
 - When the attempts have run out, we're calling `raise exc` rather than just `raise` which hides the source of the error in the traceback

Kazade:
Fixes #1149 

- Catch InternalError by default
- Re-raise original exception when retrying fails
- Log each error otherwise only the final error source will be shown
- Add an element of randomness to the retry time
- Increase the default initial wait time to 750 ms

PR checklist:
- [x] Updated CHANGELOG.md 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Added support for named class-based views to dumpurls.  Also now supports export to either json or csv
 - Added `deferred.defer_iteration_with_finalize`
 - Added `Transaction.protect_read` which prevents a specific instance being read inside a transaction.
+- Improved `djangae.utils.retry` to catch the Datastore's `InternalError`, and to better select wait times between attempts. Also improved the logging and prevented losing the source of the final exception when retrying fails.
 
 ### Bug fixes:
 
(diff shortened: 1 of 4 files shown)


adamalton:
Sorry, another post-merge review!

We're capping the back-off time at `_max_wait`, but in doing so we're removing the randomisation factor.  I think it would be better to cap the back-off time to `_max_wait` and _then_ add the randomisation factor.  I.e. this:

```python
time.sleep(min((timeout_ms + random_factor), max_wait) * 0.001)
```

would be better as this:

```python
time.sleep((min(timeout_ms, max_wait) + random_factor) * 0.001)
```

The documentation would also need tweaking accordingly.

PR #1151, opened 2018-12-10
Calculate buffer time

### LINKED ISSUE #1139: Auto-calculate buffer_time for defer_iteration_with_finalize
Issue description

In `djangae.deferred.defer_iteration_with_finalize`, there is a `buffer_time` kwarg which allows you to specify the minimum number of seconds that must be remaining in the task execution time in order for it to process another entity, rather than redeferring itself.

But rather than requiring the user to specify this buffer time, we could just calculate it automatically by setting it to the longest time that any of the entities in the shard have taken to process so far, probably plus some margin, e.g. 10%.  We would do this by just timing each iteration of the processing loop and keeping a variable with the longest time.  Then we use `longest_time * 1.1` (i.e. plus 10%) as the `buffer_time`.

Traceback or other info

* Djangae version: master@9043900ad485d84e9a2dbe3b54fcc66884bd533d

Kazade:
Fixes #1139 

- Default the `_buffer_time` arg to `defer_iteration_with_finalize` to None
- If the `_buffer_time` is None, calculate the buffer time from the longest previous iteration

Note: I haven't updated the CHANGELOG as `defer_iteration_with_finalize` is a new function. I also have been unable to think of how to write a good test for this...

PR checklist:
- [ ] Updated CHANGELOG.md 

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,3 +1,3 @@
 [flake8]
 
-max-line-length = 100
+max-line-length = 120
--- a/djangae/deferred.py
+++ b/djangae/deferred.py
@@ -19,18 +19,17 @@ This defer is an adapted version of that one, with the following changes:
 
 import copy
 import logging
+import time
 
-from django.db import models
-from django.utils import timezone
 from google.appengine.api.datastore import Delete
-from google.appengine.ext.deferred import PermanentTaskFailure  # noqa
 
 from djangae.db import transaction
 from djangae.environment import task_queue_name
 from djangae.models import DeferIterationMarker
 from djangae.processing import find_key_ranges_for_queryset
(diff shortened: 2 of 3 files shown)


adamalton:
I know this has just been merged, but seeing the notification has made me look at it and notice something: I think we should change the way that the `buffer_time_to_apply` is calculated.

Currently we're using `longest_iteration * 1.1`.  But if the iterations are very fast (e.g. 0.5 seconds) then the `buffer_time_to_apply` will end up as 0.55 seconds, which might not be long enough to redefer the task.  I think we should always add 1 second to it, as that should ensure that we've got enough time for another iteration _plus_ enough time to redefer the task afterwards.

PR #1155, opened 2019-01-04
Correct spelling mistakes

mattdoughty:


adamalton:
I liked the grater, it was good, but a bit too cheesy.

PR #1158, opened 2019-01-23
add new option remove_duplicates to ListField and RelatedListField

### LINKED ISSUE #1017: Add remove_duplicates option to ListField/RelatedListField
## Issue description

We give two main options for storing arrays of data on a model, these are `ListField` and `SetField`. The advantage of `ListField` is that it remains ordered, whereas the advantage of `SetField` is that it's not possible that it can contain duplicate entries.

However, sometimes you need order, but also to not duplicate values (in fact, that's fairly common in `ListFields`). It's simple enough to remove duplicates and retain order so I suggest we add a `remove_duplicates` keyword argument to `ListField` (defaulting to `False`) which when saving will remove any duplicated elements. 

We should also add this to `RelatedListField` if we are adding it to `ListField`

michal-macioszczyk:
Fixes #1017

- add new kwargs option "remove_duplicates" to ListField and RelatedListField
- fix some PEP-8 warnings 

PR checklist:
- [x] Updated CHANGELOG.md 

### DIFF
--- a/djangae/db/utils.py
+++ b/djangae/db/utils.py
@@ -473,3 +473,12 @@ def ensure_datetime(value):
     if isinstance(value, long):
         return datetime.fromtimestamp(value / 1e6)
     return value
+
+
+def remove_duplicates_form_list(list_to_scan):
+    remove_duplicates = []
+    for idx in xrange(len(list_to_scan)):
+        el = list_to_scan[idx]
+        if el not in remove_duplicates:
+            remove_duplicates.append(el)
+    return remove_duplicates
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - Updated `djangae.contrib.backup` to use the new export API (the existing API was deprecated in Feb 2018). This adds a dependency of `google-auth` and `google-api-python-client`,
   and also requires some manual permissions to be configured for the app service account. Existing djangae settings will be respected. Read https://cloud.google.com/datastore/docs/schedule-export for details on the new permissions required, and https://cloud.google.com/datastore/docs/export-import-entities
   for an overview including differences between the two APIs.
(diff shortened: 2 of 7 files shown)


djangae/db/utils.py
```diff
@@ -473,3 +473,12 @@ def ensure_datetime(value):
     if isinstance(value, long):
         return datetime.fromtimestamp(value / 1e6)
     return value
+
+
+def remove_duplicates_form_list(list_to_scan):
+    remove_duplicates = []
+    for idx in xrange(len(list_to_scan)):
```
davidwtbuxton: xrange has been removed in Python 3.

Why not use `for el in list_to_scan`? We don't need the index.

djangae/db/utils.py
```diff
@@ -473,3 +473,12 @@ def ensure_datetime(value):
     if isinstance(value, long):
         return datetime.fromtimestamp(value / 1e6)
     return value
+
+
+def remove_duplicates_form_list(list_to_scan):
+    remove_duplicates = []
```
davidwtbuxton: Would "unique_items" be a better name for this list?

djangae/db/utils.py
michal-macioszczyk: ah good spot ! 👍

djangae/db/utils.py
michal-macioszczyk: yes it is better name 👍

djangae/db/utils.py
```diff
@@ -473,3 +473,12 @@ def ensure_datetime(value):
     if isinstance(value, long):
         return datetime.fromtimestamp(value / 1e6)
     return value
+
+
+def remove_duplicates_form_list(list_to_scan):
```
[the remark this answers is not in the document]
michal-macioszczyk: 👍

PR #1164, opened 2019-03-08
Update "Google multiple sign-in on App Engine here" link URL

ethagnawl:
The [current link](https://p.ota.to/blog/2014/2/google-multiple-sign-in-on-app-engine/) results in a 404, so I've replaced it with a link to the Medium article which appears to be its replacement.


adamalton:
Thanks!

PR #1165, opened 2019-03-12
Python 3 Runtime (Djangae 2.x)

Kazade:
This is a WIP branch to start restructuring Djangae to be suited for the Google App Engine Python 3 Runtime.

The Datastore connector is being moved to the [django-gcloud-connectors](https://github.com/potatolondon/django-gcloud-connectors) project. This is a dependency of this branch.

Other key changes:

 - gauth, mappers and consistency contrib apps are gone
 - The test setup is now handled via tox (so no testapp)
 - memcache will need moving over to MemoryStore
 - deferred will need moving to Cloud Tasks

This is going to be a long process, at some point soon we'll fork master to `stable/1.x` and this will become the new master.


jacobg:
Hi just wondering if there's any plans to continue with this? I think there was mention that @Kazade was taking an extended leave of absence?

@adamalton

PR #1166, opened 2019-03-22
Replace yaml.load with yaml.full_load

llarco:

- yaml.load is deprecated. This PR replaces it with yaml.full_load

PR checklist:
- [N/A] Updated relevant documentation
- [N/A] Updated CHANGELOG.md 
- [N/A] Added tests for my change

### DIFF
--- a/djangae/sandbox.py
+++ b/djangae/sandbox.py
@@ -493,7 +493,7 @@ def activate(sandbox_name, add_sdk_to_path=False, new_env_vars=None, **overrides
     # when you are not running dev_appserver
     import yaml
     with open(os.path.join(project_root, 'app.yaml'), 'r') as app_yaml:
-        app_yaml = yaml.load(app_yaml)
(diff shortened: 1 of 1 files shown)


adamalton:
Thanks for submitting this.

All the tests on this fail with `AttributeError: 'module' object has no attribute 'full_load'`.  Or at least, they do on Travis.  Do the tests pass for you locally?

If so then maybe the setup on Travis is using a different version of `yaml` to the one that gets used when running the tests locally.

adamalton:
Fixes #1169.

davidwtbuxton:
If there is no need to use the full set of YAML features, we should use `yaml.safe_load()` not `yaml.full_load()`.

llarco:
Sorry for the delay! Switching it to `safe_load` worked and did not require a different version of pyyaml.

jacobg:
Potentially related:
https://github.com/potatolondon/djangae/pull/1163#discussion_r298198217

PR #1172, opened 2019-06-02
Sandbox fixes for datastore emulator and for remote shell

jacobg:
sandbox activation needs to: (1) set start api server so that datastore emulator is available during app initialization; and (2) configure remote api also so that datastore is available during app initialization. change remote shell back to sandbox configuration rather than a command, since command handling occurs after app initialization (too late in the start-up process).

- 
- 
- 

PR checklist:
- [ ] Updated CHANGELOG.md 


adamalton:
Thanks.  It looks like this just needs a couple of bits of the documentation updating (or putting back), e.g. in docs/sandbox.md  and in CHANGELOG.md.

jacobg:
Ah, I'll update the docs. Thanks!

jacobg:
Docs have been updated.

PR #1189, opened 2020-01-31
Improve locking transaction

### LINKED ISSUE #1187: contrib.locking suffers TransactionFailedErrors
Issue description

Under contention the locking app fails with TransactionFailedErrors, there are several problems with the transaction code here: https://github.com/potatolondon/djangae/blob/master/djangae/contrib/locking/models.py#L31

 - atomic() should be independent=True
 - The whole transaction should be wrapped in a try/except TransactionFailedError which should 
act as though the lock couldn't be acquired (probably just nest trans() in wrapper function with a try/except in it)
 - time.sleep should have a little randomness applied (probably between 0 and 1.0) 

Steps to reproduce

Traceback or other info

* Djangae version: 1.0.0
* Django version: 1.11
* App Engine SDK version: 1.9.54
* Using Cloud SDK: no

bogdanl:
Fixes #1187.

- Handle transaction errors when trying to acquire a lock
- Improved the locking retry countdown.

PR checklist:
- [x] Updated CHANGELOG.md 

@Kazade re #1187, not sure where you meant the independent arg should be added? `transaction.atomic` doesn't support that specific one as far as I know.

### DIFF
--- a/djangae/contrib/locking/models.py
+++ b/djangae/contrib/locking/models.py
@@ -1,6 +1,7 @@
 # STANDARD LIB
 from datetime import timedelta
 import hashlib
+import random
 import time
 
 # THRID PARTY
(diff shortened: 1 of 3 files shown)


bogdanl:
@Kazade The CI builds seem to time out, any idea why it's happening? They're also not finishing when I run them locally, but can't see how the test I added would cause this.

djangae/contrib/locking/models.py
```diff
@@ -27,28 +28,34 @@ def acquire(self, identifier, wait=True, steal_after_ms=None):
...
-                    lock.timestamp = timezone.now()
-                    lock.save()
-                    return lock
-            else:
-                return DatastoreLock.objects.create(
-                    identifier_hash=identifier_hash,
-                    identifier=identifier
-                )
+            """ Wrapper for the atomic transaction that handles transaction errors """
+            @transaction.atomic()
```
Kazade:
```suggestion
            @transaction.atomic(independent=True)
```

Kazade: https://github.com/potatolondon/djangae/blob/master/djangae/db/transaction.py#L356

Independent transactions happen outside the current transaction, which is what we want for locks

PR #1190, opened 2020-02-10
Storage backend py3

artoale:

- Rewrite the storage backend using the new python client
- bundle in the storage emulator in test runner

PR checklist:
- [x] Updated CHANGELOG.md 


pytest.ini
```diff
@@ -0,0 +1,10 @@
+[pytest]
```
artoale: This is `REALLY` useful to run tests in vscode, but we migth want to remove it?

davidwtbuxton:
Are you _really_ going to implement a cloud storage emulator? It is a pretty big API surface.

artoale:
@davidwtbuxton short answer: [yes](https://gitlab.com/potato-oss/google-cloud/gcloud-storage-emulator)

Long answer: not really. The idea is to provide _just enough_ APIs to support basic operations as required to write and run a Django Storage backend. Current (probably highly brittle) implementation supports basic operation, with many of the fields mocked out. There's no plan on implementing stuff like ACL and IAM for example, which massively simplify things

djangae/storage.py
```diff
@@ -0,0 +1,130 @@
...
+    def client(self):
+        if self._client is None:
+            self._client = _get_storage_client()
+        return self._client
+
+    @property
+    def bucket(self):
+        if not self._bucket:
+            try:
+                self._bucket = self.client.create_bucket(self._bucket_name)
```
davidwtbuxton: Ew, this is dangerous behaviour to enable by default.

davidwtbuxton:
Impressive, but mental. Now I want to do the same, write a more complete storage emulator using FastAPI for the server. And _surely_ there's something one can use to take a discovery document and generate all the schema types and the API surface.

https://www.googleapis.com/discovery/v1/apis/storage/v1/rest

Like, the server-side to what [google-api-python-client does](https://github.com/googleapis/google-api-python-client).

Also I find it frightening that this storage backend will  create a bucket for you if it doesn't exist. In my opinion creating buckets should _not_ happen by default. And it means this code assumes the service account has bucket create permissions (which is not the default).

djangae/storage.py
artoale: Fair point - tbh, the default name shouldn't be 'test-bucket' I'd imagine :)

artoale: Updated, now no bucket creation happens :)

Kazade:
OK this looks really great! Nice work!

Just a couple of things:

 - We've added a hard dependency on 'requests' - I think this is fine, I just thought it's worth making an explicit statement on here that this is what we're doing. We probably need to update setup.py though to include it

 - Do you know if this backend works with existing data/files from Djangae 1?

artoale:
> OK this looks really great! Nice work!
> 
> Just a couple of things:
> 
> * We've added a hard dependency on 'requests' - I think this is fine, I just thought it's worth making an explicit statement on here that this is what we're doing. We probably need to update setup.py though to include it

Good point! Maybe the import should be local, so that it's not required in production? We'd still need to add it to the setup.py though

> * Do you know if this backend works with existing data/files from Djangae 1?

Absolutely no idea - need to find a Djangae 1 project using them and try to access same file with the new stuff presumably. Let me know if you know of any?

artoale:
Right - @Kazade, AFAICT the datastore models are saved the same way using the python2/python3 version, so don't see an obvious reason why it shouldn't work - I guess if you're happy with the rest we might merge it, test it in the wild and see what happens :) 

Also, added the requests explicit dependency

PR #1201, opened 2020-03-04
Add target-port when starting tasks emulator

### LINKED ISSUE #1200: start_emulators should pass the --target-port to the task emulator
Currently the Cloud Tasks Emulator submits requests to port 80, which is useless for local Django development :)

Instead, start_emulators should do the following when starting the task emulator:

1. If sys.argv[1] == "runserver", parse any port override and default to port 8000
2. Else, use port 8000

artoale:
Fixes #1200 

- Add `--target-port` when starting the tasks emulator
- Allow to configure manually
- If configured automatically, try to detect from `runserver` invocation

### DIFF
--- a/djangae/sandbox.py
+++ b/djangae/sandbox.py
@@ -1,6 +1,7 @@
 import logging
 import os
 import subprocess
+import sys
 import time
 from datetime import datetime
 from urllib.error import (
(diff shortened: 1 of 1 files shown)


djangae/sandbox.py
```diff
@@ -89,9 +91,19 @@ def start_emulators(persist_data, emulators=None, storage_dir=None):
         _wait_for_datastore(DATASTORE_PORT)
 
     if "tasks" in emulators:
+        if task_target_port is None:
+            if sys.argv[1] == "runserver" and autodetect_task_port:
```
Kazade: This needs to be a bit smarter, there are other potential arguments that could be passed to runserver: https://github.com/django/django/blob/master/django/core/management/commands/runserver.py#L37

One potential avenue is to just regex each argument for IP:PORT. There's actually a regex for this here: https://github.com/django/django/blob/master/django/core/management/commands/runserver.py#L16

PR #1203, opened 2020-03-10
Don't try to use appengine credentials

Kazade:
- App Engine credentials aren't available on Python 3 so don't specify any credentials

PR checklist:
- [x] Updated CHANGELOG.md 


davidwtbuxton:
BTW the service account's application default credentials are available on the Python 3 standard runtime.  You can fetch a bearer token from the metadata service directly, or use `google.auth.default()` (with the google-auth package) which handles loading credentials from the appropriate environment.

You may need to do a bit more work if you get errors about incorrect auth scopes. More info: https://github.com/google/gae-secure-scaffold-python3/tree/master/examples/service-account-scopes

PR #1233, opened 2020-07-13
Handle task routing using defer

andreipetre:
#### 1. Default to routing tasks to their parent GAE version

Cloud Tasks run on the default GAE version, by default. I suggest we change the default behaviour to route tasks to the GAE version which created the task. This will ensure that the deferred task exists.

#### 2. Allow user to control task routing

When calling `deferred.defer()`, allow the user to control [task routing](https://cloud.google.com/tasks/docs/reference/rest/v2/AppEngineRouting) using the `_service`, `_version`, and `_instance` parameters.

### PR checklist:
- [x] Updated CHANGELOG.md 

### DIFF
--- a/djangae/tasks/deferred.py
+++ b/djangae/tasks/deferred.py
@@ -26,7 +26,7 @@ import types
 from datetime import timedelta
 from urllib.parse import unquote
 
-from djangae.environment import task_queue_name
+from djangae.environment import task_queue_name, gae_version
 from djangae.models import DeferIterationMarker
 from djangae.processing import find_key_ranges_for_queryset
 from djangae.utils import retry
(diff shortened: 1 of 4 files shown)


djangae/tasks/deferred.py
```diff
@@ -256,7 +276,8 @@ def serialize(obj, *args, **kwargs):
             'app_engine_http_request': {  # Specify the type of request.
                 'http_method': 'POST',
                 'relative_uri': deferred_handler_url,
-                'body': pickled
+                'body': pickled,
```
andreipetre, line 279: Should this also contain `headers`?

Kazade: Possibly, but at can do that as another PR, this looks great!

andreipetre: fyi, I opened #1234 for this.

PR #1237, opened 2020-07-24
Oauth contrib auth

marcoazza:

- 
- 
- 

PR checklist:
- [ ] Updated CHANGELOG.md 


Kazade:
Thanks Marco, I'm just going to make some notes for myself here that I'll pick up over the next few days (just me thinking aloud!):

 - We're probably OK to depend on the AbstractBaseUser from django.contrib.auth, in fact, for compatibility we should probably continue to do that
 - Same for utility functions (e.g. _get_backends, validators)
 - Probably need to restore default_app_config to `__init__.py`
 - Should move OAuthUserSession creation to the oauth2callback, and authenticate of the oauthbackend should check that (or indeed, be passed it)
 - Need to consider what happens when you have multiple backends activated (e.g. IAP)
 - Need to take another look at the credential stuff that David Buxton did recently
 - Need to think about how to include the e2e test

Kazade:
Hmm, I've spent a good while looking at this and I can't figure out how you got a client ID from the google.auth.default credentials call? For now I'm going to revert back to storing a client ID + secret in settings until I can figure it out

PR #1242, opened 2020-08-07
Oauth contrib auth

Kazade:
- Adds `djangae.contrib.googleauth` as a replacement to the old `djangae.contrib.gauth`
- Adds an oauth2 backend for authentication with Google's oauth2 auth service
- Adds an IAP backend for authentication with Internet Aware Proxy
- Adds a permission system that doesn't rely on ContentTypes due to issues they cause with a non-relational datastore
- Adds djangae.credentials as a new package

Still to do:
 - Write more tests for OAuth
 - Store the requested scopes on the OAuthUserSession
 - Implement the IAP backend
 - Determine if IAP and Oauth user IDs are the same for the same user, if so, update google_oauth_id -> google_user_id

PR checklist:
- [x] Updated CHANGELOG.md 


Kazade:
OK a bit of an update this:

 - IAP authentication works, but we're not yet using JWT 
 - OAuth now works, and you can actually combine IAP with Oauth for permission granting, so that's nice
 - There's still not a solution for the old `is_appengine_admin` thing from Python 2 days. I'm inclined to do that in a separate branch
 - Tests need fixing and more writing, and things need documenting, but it's almost there

Kazade:
@artoale Could you give this a review, and then "Squash & Merge"

I'll create follow up tickets for the remaining tasks :)

artoale (commented):
Looking great - nice one @Kazade !

Left a couple of small comments in the review - one question, are we considering [IAP signed headers verification](https://cloud.google.com/iap/docs/signed-headers-howto)?

PR #1245, opened 2020-08-20
Fix error classes

andreipetre:
### Summary of changes proposed in this Pull Request:
- fix error classes not inheriting from `Exception`

### PR checklist:
- [x] Updated CHANGELOG.md 

### DIFF
--- a/djangae/utils.py
+++ b/djangae/utils.py
@@ -114,14 +114,14 @@ def retry(func, *args, **kwargs):
         # If gcloudc is available, make sure we catch its TransactionFailedError
         from gcloudc.db.transaction import TransactionFailedError
     except ImportError:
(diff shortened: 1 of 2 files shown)


djangae/utils.py
```diff
@@ -114,14 +114,14 @@ def retry(func, *args, **kwargs):
         # If gcloudc is available, make sure we catch its TransactionFailedError
         from gcloudc.db.transaction import TransactionFailedError
     except ImportError:
-        class TransactionFailedError:
+        class TransactionFailedError(Exception):
```
andreipetre, line 117: Otherwise we get the following error:

```
TypeError: catching classes that do not inherit from BaseException is not allowed
```

PR #1260, opened 2020-10-16
djangae.contrib.search

Kazade:
- Implements a basic replacement for the App Engine Search API as a Django contrib app
- Provides a similar API to create Indexes and Documents
- Provides Django queryset integration for searching for model instances

PR checklist:
- [x] Updated CHANGELOG.md 


tox.ini
```diff
@@ -15,7 +15,7 @@ commands =
     pip install requests-oauthlib
     pip install google-auth-oauthlib
     pip install google-cloud-tasks<2.0.0
-    pip install selenium
+    pip install selenium nltk
```
sleepyjames, line 18: nltk is unused, can be removed.

sleepyjames (changes requested):
As discussed elsewhere we can revisit how indexing is triggered in that:

-  Performing this all within the model's `save` method will likely cause problems once the number of fields indexed and their content increases beyond a relatively small amount.
-  De-coupling indexing from save has the advantage that we can do it async.

sleepyjames (approved):
Awesome, thanks @Kazade

PR #1262, opened 2020-10-19
Generate username function now being used in OAuth flow

purvitrivedi:

- `_generate_unused_username` now being used in `oauth2.py`
- Profile object does not contain username - only userId, full name, given name, family name, image URL and email, so we won't create add char field `profile_username` to `AbstractGoogleUser`

PR checklist: 

### DIFF
--- a/djangae/contrib/googleauth/backends/oauth2.py
+++ b/djangae/contrib/googleauth/backends/oauth2.py
@@ -5,9 +5,8 @@ from ..models import (
     UserManager,
     UserPermission,
 )
-from . import _find_atomic_decorator
+from . import _find_atomic_decorator, _generate_unused_username
 from .base import BaseBackend
-from .iap import _generate_unused_username
 
 
 class OAuthBackend(BaseBackend):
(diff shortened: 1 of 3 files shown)


djangae/contrib/googleauth/backends/oauth2.py
```diff
@@ -7,6 +7,7 @@
 )
 from . import _find_atomic_decorator
 from .base import BaseBackend
+from .iap import _generate_unused_username
```
Kazade, line 10: Could you move the `_generate_unused_username` function into `backends/__init__.py` and then import from there in both iap.py and here? (e.g. `from . import _generate_unused_username`)

Kazade (changes requested):

djangae/contrib/googleauth/backends/oauth2.py
purvitrivedi: Done!

flea89:
Fixed by https://github.com/potatolondon/djangae/pull/1262

PR #1281, opened 2020-11-09
Search improvements

Kazade:
- Fixes search eventual consistency issues on the older datastore
- Fixes handling of full stops in strings

PR checklist:
- [x] Updated CHANGELOG.md 


jacobg:
@Kazade Thanks for this! When you say "Fixes search eventual consistency issues on the older datastore", do you mean djangae v2 works on the older datastore? The django-gcloud-connectors repo documents that only Datastore in Firebase Mode is supported.

Kazade:
Well, there's "supported" and there's "supported" :)

The old datastore isn't really catered for - we still don't have any patches over eventual consistency. However, if you accept that your queries won't be consistent, and you're limited in transactions then for now* things should work. This just happened to be an easy win to make it a key-lookup, rather than a query.

*This won't be the case when we switch to using Django's atomic() decorator which will probably break everything if you're not on Firestore in Datastore mode.

jacobg:
Thanks very much @Kazade. Is there a time frame on atomic decorator? Is it being intentionally delayed until Google migrates existing datastores to Firestore in Datastore mode?

sleepyjames (commented):
This appears to break searching for acronyms that use periods, for example, given the doc:

```python
class Doc(Document):
    text = fields.TextField()
```

Searching using hiphens works:

```python
> index.add(Doc(text="foo-bar-baz"))
> len(list(index.search("foo-bar-baz", Doc)))
1
```
 but using period doesn't:
```python
> index.add(Doc(text="foo.bar.baz"))
> list(index.search("foo.bar.baz", Doc))
[]
```

Kazade:
Thanks @sleepyjames - just pushed a fix :)

sleepyjames:
LGTM 

And yes "foo.bar.baz" it not an acronym 🤦

PR #1283, opened 2020-11-10
1282 sandbox

### LINKED ISSUE #1282: Check if emulators are already running before spawning a new instance.
The sandbox `start_emulators` function attempts to start new instances of the emulators despite the fact that one could already be running.

Let's check if one already exists before trying to create a new one.

* Djangae version: x.x.x / master@commit
* Django version: x.x.x
* App Engine SDK version: 1.9.xx
* Using Cloud SDK: yes/no

flea89:
Fixes #1282

- when starting a new emulator, use the first available port if not pass explicitly 
- Raise an exception if the port is passed and not available

PR checklist:
- [ ] Updated CHANGELOG.md 

### DIFF
--- a/djangae/sandbox.py
+++ b/djangae/sandbox.py
@@ -14,11 +14,15 @@ from urllib.request import urlopen
 from django.utils.autoreload import DJANGO_AUTORELOAD_ENV
 
 from djangae.environment import get_application_root
+from djangae.utils import get_next_available_port, port_is_open
 
 _ACTIVE_EMULATORS = {}
 _ALL_EMULATORS = ("datastore", "tasks", "storage")
 
 _DJANGO_DEFAULT_PORT = 8000
+
+SERVICE_HOST = "127.0.0.1"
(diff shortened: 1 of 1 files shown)


sleepyjames:
Thanks @flea89, I have a A few questions:

- We're now overriding any value set with the environment variable `DATASTORE_EMULATOR_HOST` completely. I'm not sure we *should* be doing this but perhaps we can output a warning?
- This also now compounds the issue where the datastore emulator is not shutdown when the development server exits, so maybe we should fix that first?
- Should we now provide a method for a user to pass in the desired ports now or instead use any environment variables set as the starting point.

sleepyjames:
From offline discussion with @flea89  

- When starting the emulator the port is **not** taken from an environment variables so we're safe to set this arbitrarily.
- Can tackle how we (maybe) pass in a host/port via `manage.py` elsewhere, it's out of scope here.
- One part of the shutdown issue mentioned is being handled in https://github.com/potatolondon/djangae/pull/1286

PR #1284, opened 2020-11-11
Add the ability to order by a document field

Kazade:
- Adds the ability to pass an order_by field to index.search()

PR checklist:
- [x] Updated CHANGELOG.md 

### DIFF
--- a/djangae/contrib/search/index.py
+++ b/djangae/contrib/search/index.py
@@ -201,8 +201,10 @@ class Index(object):
             match_all=match_all,
         )[:limit]
 
+        doc_instance = document_class()
+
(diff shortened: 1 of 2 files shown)


djangae/contrib/search/index.py
```diff
@@ -200,12 +201,18 @@ def search(
             match_all=match_all,
         )[:limit]
 
+        def get_field_value(field_name, record):
+            field = document_class().get_field(field_name)
```
sleepyjames, line 205: Do we need to call `document_class()` each time here? This function can be called many times in the block below, can we get away with creating a single instance then calling `get_field` on the same one?

Issue #1301, opened 2020-11-25
Authentication redirect


flea89:
Authentication redirects break when starting the OAuth flow starting from `myversion-dot-app...` and `myversion` is set to default.

* Djangae version: x.x.x / master@commit
* Django version: x.x.x
* App Engine SDK version: 1.9.xx
* Using Cloud SDK: yes/no

PR #1302, opened 2020-11-25
Use hostname instead of version for OAuth flow redirects #1301

flea89:
Use hostname instead of version for OAuth flow redirects #1301

PR checklist:
- [ ] Updated CHANGELOG.md 


flea89:
@Kazade FYI I've renamed the setting

PR #1304, opened 2020-11-25
Fix oauth expiry

### LINKED ISSUE #1303: Authentication middleware shouldn't redirect to oauth
When someone logs in with oauth, we have two different sessions in play:

1. The Django session, which directly controls whether someone is logged in or not
2. The OAuth session

Both sessions expire, but currently the behaviour is to artificially expire the Django session when OAuth session expires. This likely isn't the correct behaviour. It comes down to a question of authentication vs authorization.

When we log in with the OAuth2 backend, we request access to the profile scope to *identify and authenticate* the user. If that OAuth session expires, it doesn't matter. If however we want an additional scope to say, update a Google Sheet, that would be *authorization* and we'd need a current oauth session for that to work.

Effectively, `@login_required` should only trigger an oauth flow if the user isn't logged in with Django, `@oauth_scopes_required` should trigger an oauth flow if we don't currently have a valid oauth session with access granted to the specified scopes.

We shouldn't log out the user if the oauth session expires, but some users may want to enforce that behaviour for tighter integration with the oauth2 session. So I'm proposing the following:

 - Adding a setting: `GOOGLEAUTH_LINK_OAUTH_SESSION_EXPIRY` that defaults to `False`
 - If True, the user will be logged out and redirected through the oauth flow when their oauth session expires (current behaviour)
 - If False, the oauth session will not log out a user from their Django session (but they will lose access to views that require a valid oauth session)

Kazade:
Fixes #1303 

- By default, don't log out a Django session just because the oauth session became invalid
- Make this configurable via the `GOOGLEAUTH_LINK_OAUTH_SESSION_EXPIRY` setting

PR checklist:
- [x] Updated CHANGELOG.md 

### DIFF
--- a/docs/googleauth.md
+++ b/docs/googleauth.md
@@ -113,7 +113,7 @@ required to configure a list of Authorised domains and OAuth redirects.
 ## Linking OAuth & Django Session Expiry
 
 By default, an oauth session expiring doesn't force log-out the user from their Django session. If however you want to *require* that a Django session
-expires when the oauth session expires, you can do so by setting the `GOOGLEAUTH_LINK_OAUTH_SESSION_EXPIRY` setting to `True`.
+expires when the oauth session expires, you can do so by setting the `GOOGLEAUTH_LINK_OAUTH_SESSION_EXPIRY` setting to `True`. This will redirect the user back through the oauth flow (although normally transparently).
 
 ## Handling App Engine Versioning
 
--- a/docs/googleauth.md
+++ b/docs/googleauth.md
@@ -107,11 +107,18 @@ GOOGLEAUTH_CLIENT_SECRET = "..."
 
 You must generate these values in the Google Cloud Console.
 
-When you configure OAuth Consent Screen and Credentials on Google Cloud Platform you are 
+When you configure OAuth Consent Screen and Credentials on Google Cloud Platform you are
 required to configure a list of Authorised domains and OAuth redirects.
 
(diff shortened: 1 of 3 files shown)


docs/googleauth.md
```diff
@@ -107,11 +107,18 @@ GOOGLEAUTH_CLIENT_SECRET = "..."
...
 You must generate these values in the Google Cloud Console.
 
-When you configure OAuth Consent Screen and Credentials on Google Cloud Platform you are 
+When you configure OAuth Consent Screen and Credentials on Google Cloud Platform you are
 required to configure a list of Authorised domains and OAuth redirects.
 
+## Linking OAuth & Django Session Expiry
+
+By default, an oauth session expiring doesn't force log-out the user from their Django session. If however you want to *require* that a Django session
+expires when the oauth session expires, you can do so by setting the `GOOGLEAUTH_LINK_OAUTH_SESSION_EXPIRY` setting to `True`.
```
flea89, line 116: This might be a bit misleading. What really happens when you have `GOOGLEAUTH_LINK_OAUTH_SESSION_EXPIRY = True` is that the user is forced through the authorisation flow again rather than really being logged out. ( and as it stand right now, it goes "transparently" now, doesn't show the prompt if a user session exists.
I'm not sure if you want to keep that behavior or not.
I feel that certainly useful in case of `oauth_scopes_required`, but not here?

PR #1313, opened 2020-12-10
Re-raise errors from deferred iteration shards if it's the first-iter…

### LINKED ISSUE #1312: Defer iteration should retry the current task if it fails on the first iteration
Currently if a shard hits an error, the shard is redeferred to continue where it was. This can give a false impression in the Cloud Tasks dashboard though as the retry count will be zero.

Instead, if we hit an error, we should redefer from where we were, unless we're on the first iteration when we should instead raise a SingularTaskFailed (or whatever is equivalent in the new tasks API)

Kazade:
…ation

Fixes #1312 

- If a deferred iteration shard fails on the first iteration, it now just re-raises the error to retry the same task, rather than deferring a new one.

PR checklist:
- [ ] Updated CHANGELOG.md 

### DIFF
--- a/djangae/tasks/deferred.py
+++ b/djangae/tasks/deferred.py
@@ -380,6 +380,8 @@ def _process_shard(marker_id, shard_number, model, query, callback, finalize, ar
         )
         return
 
+    first_iteration = True
+
     try:
         qs = model.objects.all()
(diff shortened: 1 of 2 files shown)


flea89:
LGTM!

PR #1314, opened 2020-12-10
Fix issue where entities may be skipped in defer_iteration_with_finalize

Kazade:
- Fixes a bug which could cause instances to be skipped while iterating with defer_iteration_with_finalize

PR checklist:
- [ ] Updated CHANGELOG.md 

### DIFF
--- a/djangae/tasks/deferred.py
+++ b/djangae/tasks/deferred.py
@@ -501,8 +501,7 @@ def _generate_shards(
         if end:
             filter_kwargs["pk__lt"] = end
 
-        # calling order_by with no args to clear any pre-existing ordering (e.g. from Meta.ordering)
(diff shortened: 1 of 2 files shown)


gglynne (commented):
`test_shard_iterated_in_order` seems to pass both with and 
without the fixes in `djangae/tasks/deferred.py` lines 388/391.
my testing locally shows the default ordering is 'pk' (`__key__`?),
does this fix work differently for larger data sets?

Kazade:
OK turns out this is really really hard to properly test in a test. 

If you provide no sort order to a query on the production datastore it returns them in some pseudo random order (docs say in the "order they were retrieved from Datastore mode" whatever that means), on the Datastore emulator though they return in key order. I think what we have is enough, it definitely fixes the bug!

PR #1321, opened 2021-01-18
Move everything into process_request

### LINKED ISSUE #1320: AuthenticationMiddleware should override process_request, not process_view
The background is that we needed the `oauth_login` view to be exempt from the `AuthenticationMiddleware`, and `process_view` allowed us to inspect the view function.

The problem is that if you have a middleware that requires a user to be logged in, process_view happens to late - after all the other middleware has run.

The fix is to remove `@auth_middleware_exempt`, move the code back to process_request, and then specifically check for the `oauth_login` view.

Kazade:
Fixes #1320 

Went a slightly different way than I described in the issue, and instead just moved the code from process_view -> process_request and used resolver_match to look up the handling
view.

- Perform authentication in process_request, instead of process_view

PR checklist:
- [x] Updated CHANGELOG.md 

### DIFF
--- a/djangae/contrib/googleauth/middleware.py
+++ b/djangae/contrib/googleauth/middleware.py
@@ -2,6 +2,7 @@
 import hashlib
 import logging
 import os
+
 from django import forms
 from django.conf import settings
 from django.contrib.auth import (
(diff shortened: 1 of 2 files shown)


flea89:
LGTM!
Merged!

